PR #2, opened 2020-07-13
Bootstrap the repo

amwat:
k8s.io/test-infra/kubetest2 -> sigs.k8s.io/kubetest2

/cc @BenTheElder


go.mod
```diff
@@ -0,0 +1,14 @@
+module sigs.k8s.io/kubetest2
+
+go 1.14
+
+require (
+	github.com/aws/aws-k8s-tester v1.0.0
```
BenTheElder, line 6: we should get this into a sub-module (just containing a tester?) for sure.

Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: install
+install:
+	go install .
```
BenTheElder, line 3: force module mode?

amwat: done

go.mod
amwat: you mean get a different go.mod for the eks deployer right?
i will probably do that later with the correct naming sorted out.

cofyc:
/lgtm

ready to merge?

PR #3, opened 2020-07-15
gke: support acquiring project from boskos 

cofyc:
I am planning to use `kubetest2` in the prow environment.  Boskos support is required.

It includes:

- move boskos to a separate pkg
- support boskos in kubetest2-gke
- register klog flags in kubetest2-gke


cofyc:
@amwat rebased

> Also, we don't have CI for this currently, can you verify it by running locally once?

I've verified it locally. Next, we can add some basic CI jobs for gke/gce/eks.

PR #7, opened 2020-07-28
Port over kind's go_container.sh

amwat:
inspired from https://github.com/kubernetes-sigs/kind/blob/master/hack/

this will make local builds much more hermetic and updating go version very straightforward.

/cc @BenTheElder


amwat:
https://github.com/kubernetes/test-infra/pull/18493 to add dind support for the e2e job.
plus we'll get to run the check on this PR itself.

hack/go_container.sh
```diff
@@ -0,0 +1,96 @@
...
+# get the repo root for defaulting OUT_DIR and SOURCE_DIR
+REPO_ROOT="${REPO_ROOT:-$(git rev-parse --show-toplevel)}"
+# output directory, will be mounted to /out, defaults to /bin in REPO_ROOT
+OUT_DIR="${OUT_DIR:-${REPO_ROOT}/bin}"
+# source directory, will be mounted to /src, defaults to current directory
+SOURCE_DIR="${SOURCE_DIR:-$(pwd -P)}"
+# default to disabling CGO for easier reproducible builds and cross compilation
+export CGO_ENABLED="${CGO_ENABLED:-0}"
+# the container image, by default a recent official golang image
+GOIMAGE="${GOIMAGE:-golang:1.15beta1}"
```
BenTheElder, line 33: probably want rc1 now

hack/tools/go.sum
```diff
@@ -0,0 +1,585 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
BenTheElder, line 1: not sure if you did, but you'll want to `go mod tidy`,  it also minimizes the sum file a bit

hack/tools/tools.go
```diff
@@ -0,0 +1,16 @@
...
+package tools
+
+
+import (
+	// linter(s)
+	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
+
+	// test runner
+	_ "gotest.tools/gotestsum"
+)
```
BenTheElder, line 16: is this gofmt-ed?

hack/go_container.sh
amwat: done

hack/tools/go.sum
amwat: done

hack/tools/tools.go
amwat: nice catch, seems lint wasn't running as expected. should be fixed now.

PR #9, opened 2020-07-30
Move out testers to separate binaries.

amwat:
Made minimal modifications to the tester logic to remove old dependencies and still keep it working.
Will be followed-up by making them better suited for the new flow.

/assign @BenTheElder 
/cc @michaelmdresser


pkg/testers/ginkgo/ginkgo.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+const (
+	binary = "ginkgo" // TODO(RonWeber): Actually find these binaries.
+)
+
+// Fixing this path temporarily for local testing
+// TODO(amwat): implement actual logic
+var (
+	e2eTestPath = filepath.Join("_artifacts", "kubernetes", "test", "bin", "e2e.test")
```
michaelmdresser, line 40: This path was already broken, so I think this is fine for now.

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -0,0 +1,127 @@
...
+		"--"}, e2eTestArgs...)
+
+	log.Printf("Running ginkgo test as %s %+v", binary, ginkgoArgs)
+	cmd := exec.Command(binary, ginkgoArgs...)
+	exec.InheritOutput(cmd)
+	return cmd.Run()
+}
+
+func (t *Tester) pretestSetup() error {
+	// TODO(amwat): pass this from kubetest2
```
michaelmdresser, line 76: How will getting the kubeconfig from the deployer eventually work?

BenTheElder:
I think you want to name all binary packages so that `go get sigs.k8s.io/kubetest2/...` just works without needing make etc.

pkg/testers/ginkgo/ginkgo.go
amwat: resolved, kubetest2 core will now pass $KUBECONFIG if deployers provide it.

amwat:
They are? except you need an additional kubetest2 for the main repo as well `go get sigs.k8s.io/kubetest2/kubetest2-tester-ginkgo` should work.

pkg/app/cmd.go
```diff
@@ -123,20 +128,6 @@ func runE(
 		}
 	}
 
-	// instantiate the tester if testing was specified
-	// NOTE: we need to do this before building etc, so we can fail fast on
-	// invalid options to the tester
-	var tester types.Tester
-	if newTester != nil {
-		t, err := newTester(opts, testerArgs, deployer)
```
michaelmdresser, line 131: Removing this removes an opportunity for a runtime check of tester options before it is actually executed. This could put the user in a situation where the deployer goes through build and up and _then_ the tester errors because of a bad flag/option. I'm specifically thinking of the artifacts dir change in #8, but this could impact other future flags as well.

michaelmdresser: Not saying it necessarily has to be checked before the deployer is called, but it would help the UX.

amwat: Yes, I think that is an expected side effect. I'm planning to make `--help` a requirement for all testers, in which case we could overload it's functionality to also ensure all flags are parsed.

amwat: Okay, I added a fix for this behavior since this seems pretty important UX.

PR #11, opened 2020-07-31
Default artifacts is absolute path and Ginkgo tester expands KUBECONFIG to absolute path

michaelmdresser:
Odd behavior was encountered when running the ginkgo tester with non-absolute paths for ARTIFACTS after running the GCE deployer. Observed behavior shows that when the ginkgo executable is called with a non-absolute path for KUBECONFIG it errors. Because the GCE deployer merely adds paths on top of ARTIFACTS, it was returning a KUBECONFIG (sent to the tester) of ARTIFACTS/kubetest2-kubeconfig which was a relative path because ARTIFACTS is, by default, "_artifacts".

This PR addresses the problem at 2 levels: it changes the default artifacts dir to have an absolute path and then ensures an absolute path for the kubeconfig passed to the ginkgo executable. While this problem could be addressed by only ensuring the kubeconfig path is absolute in the ginkgo tester, it is also better to have a good default for artifacts in case of future problems.

This is a post #9 version of #8.

### DIFF
--- a/pkg/app/cmd.go
+++ b/pkg/app/cmd.go
@@ -20,7 +20,10 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
+
+	"k8s.io/klog"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -164,12 +167,23 @@ func runE(
 }
 
 // the default is $ARTIFACTS if set, otherwise ./_artifacts
-func defaultArtifactsDir() string {
+// constructed as an absolute path to help the ginkgo tester because
+// for some reason it needs an absolute path to the kubeconfig
+func defaultArtifactsDir() (string, error) {
 	path, set := os.LookupEnv("ARTIFACTS")
 	if set {
-		return path
+		absPath, err := filepath.Abs(path)
+		if err != nil {
+			return "", fmt.Errorf("failed to convert filepath from $ARTIFACTS (%s) to absolute path: %s", path, err)
+		}
+		return absPath, nil
 	}
-	return "./_artifacts"
+
+	absPath, err := filepath.Abs("_artifacts")
+	if err != nil {
+		return "", fmt.Errorf("when constructing default artifacts dir, failed to get absolute path: %s", err)
+	}
+	return absPath, nil
 }
 
 // splitArgs splits args into deployerArgs and testerArgs at the first bare `--`
@@ -205,7 +219,12 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.up, "up", false, "provision the test cluster")
 	flags.BoolVar(&o.down, "down", false, "tear down the test cluster")
 	flags.StringVar(&o.test, "test", "", "test type to run, if unset no tests will run")
-	flags.StringVar(&o.artifacts, "artifacts", defaultArtifactsDir(), `directory to put artifacts, defaulting to "${ARTIFACTS:-./_artifacts}"`)
+
+	defaultArtifacts, err := defaultArtifactsDir()
+	if err != nil {
+		klog.Fatalf("failed to get default artifacts directory: %s", err)
+	}
+	flags.StringVar(&o.artifacts, "artifacts", defaultArtifacts, `directory to put artifacts, defaulting to "${ARTIFACTS:-./_artifacts}". If using the ginkgo tester, this must be an absolute path.`)
 }
 
 // assert that options implements deployer options
--- a/pkg/testers/ginkgo/ginkgo.go
+++ b/pkg/testers/ginkgo/ginkgo.go
@@ -70,6 +70,20 @@ func (t *Tester) Test() error {
 
 func (t *Tester) pretestSetup() error {
 	if config := os.Getenv("KUBECONFIG"); config != "" {
+		// The ginkgo tester errors out if the kubeconfig provided
+		// is not an absolute path, likely because ginkgo changes its
+		// working directory while executing. To get around this problem
+		// we can manually edit the provided KUBECONFIG to ensure a
+		// successful run.
+		if !filepath.IsAbs(config) {
+			newKubeconfig, err := filepath.Abs(config)
+			if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/app/cmd.go
```diff
@@ -164,12 +165,19 @@ func runE(
...
-		return path
+		return path, nil
 	}
-	return "./_artifacts"
+
+	workDir, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("when constructing default artifacts dir, failed to get working directory: %s", err)
+	}
+	return filepath.Join(workDir, "_artifacts"), nil
```
amwat, line 180: filepath.Abs ?

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -70,6 +70,21 @@ func (t *Tester) Test() error {
...
+		// The ginkgo tester errors out if the kubeconfig provided
+		// is not an absolute path, likely because ginkgo changes its
+		// working directory while executing. To get around this problem
+		// we can manually edit the provided KUBECONFIG to ensure a
+		// successful run.
+		if !filepath.IsAbs(config) {
+			workDir, err := os.Getwd()
+			if err != nil {
+				return fmt.Errorf("failed to get working dir to set absolute kubeconfig path: %s", err)
+			}
```
amwat, line 82: same here

pkg/app/cmd.go
```diff
@@ -205,7 +213,12 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.up, "up", false, "provision the test cluster")
 	flags.BoolVar(&o.down, "down", false, "tear down the test cluster")
 	flags.StringVar(&o.test, "test", "", "test type to run, if unset no tests will run")
-	flags.StringVar(&o.artifacts, "artifacts", defaultArtifactsDir(), `directory to put artifacts, defaulting to "${ARTIFACTS:-./_artifacts}"`)
+
+	defaultArtifacts, err := defaultArtifactsDir()
+	if err != nil {
+		panic(err)
```
amwat, line 219: we shouldnt panic.
maybe klog.fatalf?

pkg/app/cmd.go
michaelmdresser: Didn't know that was a thing, done.

pkg/testers/ginkgo/ginkgo.go
michaelmdresser: Done.

pkg/app/cmd.go
michaelmdresser: Done.

pkg/app/cmd.go
```diff
@@ -164,12 +167,19 @@ func runE(
...
 
 // the default is $ARTIFACTS if set, otherwise ./_artifacts
-func defaultArtifactsDir() string {
+// constructed as an absolute path to help the ginkgo tester because
+// for some reason it needs an absolute path to the kubeconfig
+func defaultArtifactsDir() (string, error) {
 	path, set := os.LookupEnv("ARTIFACTS")
 	if set {
-		return path
+		return path, nil
```
amwat, line 175: nit: do we also want to abspath this?

michaelmdresser: Done, also caused me to notice and fix a slight bug with the default.

BenTheElder:
I'm not sure we should be mutating the kubeconfig value.
Are we mutating the current working directory?
Who is setting a relative path? Why isn't it absolute at this level?

michaelmdresser:
We aren't mutating the working directory. The artifacts directory was defaulting to a relative path and could be also be manually set to a relative path with `--artifacts`. The GCE deployer would put its kubeconfig in the artifacts directory and return a relative path to it for the tester to consume. The ginkgo tester would therefore be configured to use this relative path to the kubeconfig, which fails. Changing it to an absolute path solves the failure. I believe ginkgo is doing some sort of working directory modification.

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -70,6 +70,20 @@ func (t *Tester) Test() error {
 
 func (t *Tester) pretestSetup() error {
 	if config := os.Getenv("KUBECONFIG"); config != "" {
+		// The ginkgo tester errors out if the kubeconfig provided
```
amwat, line 73: now that artifacts is relative and all deployers expose absolute paths for their generated kubeconfigs this section is probably no longer needed ?

amwat: FYI i also suspect this bug is most likely because of passing the kubeconfig path directly as strings (line 55) which could potentially contain shell specific characters like `~` for `$HOME`
we can probably execute it as `sh -c` and drop the equals while assigning flag values, just separated args should also work.

michaelmdresser: Yeah, based on current code paths it isn't necessary in the tester itself any more, but we did talk about enforcing it at more boundaries. I'm happy to go either way with it.

As for the bug, I don't think its shell-specific chars (could be wrong) because I remember seeing in the test output a pure relative path `_artifacts/kubetest2-kubeconfig` being passed to ginkgo and it failing.

amwat: okay let's drop the tester changes if the artifacts change was sufficient for the gce deployer.
we can consider adding it back later if we run into issues again.

michaelmdresser: Actually, looking at it again. The default dir will work fine (enforced abs) but if the user passes a relative path for `--artifacts` and also uses the ginkgo tester (thus not obeying the updated usage message) kt2 will not error until after Up(), when it calls the tester. The error will also be somewhat obscure to someone who hasn't gone through this issue.

To fix it I think we either enforce it in the ginkgo tester (as we do in the current code) or in the deployer, which is, in my opinion, inappropriate because it only matters to the tester, not the deployer.

amwat: I guess we could still validate and expand user passed flag value.
but let's go forward with this current approach since it's fixing a broken behavior.

note: that eventually we'll be having cases where the kubeconfigpath is actually a comma separated list.

PR #12, opened 2020-07-31
[Proposal] Multi project multi cluster proposal for GKE deployer

albertomilan:
--project and --cluster-name parameters have been modified to accept a list of projects and cluster. This change respects previous functionality with only one project and one cluster. On multi-project cases, cluster names do have to end with "-X", being X a 0-based index to which project they should belong. No other dash character is allowed in the cluster name for parsing purposes and the final cluster name will filter out this indexing postfix.

--projects-requested=int has been included to request N projects from Boskos

Internal data structures keep track of the relations between projects and clusters and the difference instance groups

Network parameter is kept as a single element. On one project cases, it's the network used and on multi-project cases, it is the name of the Shared VPC network that will be created on the first project and to which the rest of the projects will be linked with.

 Changed default environment from 'staging' to 'prod'

Succesful regression test performed:
$ go run ~/go/src/github.com/kubernetes-sigs/kubetest2/kubetest2-gke --up --down --num-nodes=1 --project=test-project --cluster-name=testcluster --zone=us-central1-a --network=test-network


amwat:
dupe of https://github.com/kubernetes-sigs/kubetest2/pull/14 ?

chizhg:
/close
In favor of #14

PR #13, opened 2020-08-03
kubetest2 gke retries creating the cluster if it fails due to specific reasons

chizhg:
1. As we synced offline, `kubetest2 gke` supports retrying cluster creation if it fails due to some reasons that cannot be prevented in advance, e.g. quota running out. This is achieved by adding an extra `back-regions` flag which will enable the retry logic if it's not empty. IMO this approach is better than supporting `region` to be set as `auto` since it allows users to restrict which region(s) they want the cluster to be in, and it's also backward compatible.

2. Add a new flag `require-gcp-ssh-key` to allow the GCP SSH key to be empty, with which we can get rid of the hack in Knative - https://github.com/knative/test-infra/blob/master/scripts/e2e-tests.sh#L229-L233 (we are still using `kubetest` in Knative but will be likely switching to `kubetest2` soon).

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/commandutils.go
+++ b/kubetest2-gke/deployer/commandutils.go
@@ -45,7 +45,6 @@ func (d *deployer) prepareGcpIfNeeded() error {
 		return fmt.Errorf("--environment must be one of {test,staging,staging2,prod} or match %v, found %q", urlRe, env)
 	}
 
-	//TODO(RonWeber): boskos
 	if err := os.Setenv("CLOUDSDK_CORE_PRINT_UNHANDLED_TRACEBACKS", "1"); err != nil {
 		return fmt.Errorf("could not set CLOUDSDK_CORE_PRINT_UNHANDLED_TRACEBACKS=1: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


amwat:
We also need to handle zonal clusters.

I'd like to reconsider the use case for this functionality.
the references issues should be solved in different ways I think.

1) Stockout isn't something we want to hide under the covers (i.e. jobs failing and requiring reconfiguring the prowjobs seems acceptable in just adverse cases.)
2) Quota issues should definitely not be hidden by retries (this means that the job isn't configured with the right region in the first place and will always keep on failing the first time and go on to backup regions)

3) Different versions being available in different regions should be fixed by improving the "extract" behavior where the latest available gke version is determined by  using the https://cloud.google.com/sdk/gcloud/reference/container/get-server-config API and not hardcoding a version in the job config.

chizhg:
> We also need to handle zonal clusters.
> 
> I'd like to reconsider the use case for this functionality.
> the references issues should be solved in different ways I think.
> 
> 1. Stockout isn't something we want to hide under the covers (i.e. jobs failing and requiring reconfiguring the prowjobs seems acceptable in just adverse cases.)
> 2. Quota issues should definitely not be hidden by retries (this means that the job isn't configured with the right region in the first place and will always keep on failing the first time and go on to backup regions)
> 3. Different versions being available in different regions should be fixed by improving the "extract" behavior where the latest available gke version is determined by  using the https://cloud.google.com/sdk/gcloud/reference/container/get-server-config API and not hardcoding a version in the job config.

The stockout issue, which corresponds to `does not have enough resources available to fulfill`, is not a configuration error: the number of machines in a region (or a data center) is limited, and the machines will kept being acquired and released by all users from the Internet, so the number of available machines is dynamic and any region can possibly run into stockout issues.

But I agree the other issues are preventable without blindly retrying, and since some `kubetest2` users might want the clusters to be created in a fixed region (e.g. for the multi-cluster profile), so I'll close this PR for now and probably will handle this in the place where we run `kubetest2` commands.

PR #14, opened 2020-08-03
[Proposal] Multi project multi cluster proposal for GKE deployer

chizhg:
--project and --cluster-name parameters have been modified to accept a list of projects and cluster. This change respects previous functionality with only one project and one cluster. On multi-project cases, cluster names do have to end with "-X", being X a 0-based index to which project they should belong. No other dash character is allowed in the cluster name for parsing purposes and the final cluster name will filter out this indexing postfix.

--projects-requested=int has been included to request N projects from Boskos

Internal data structures keep track of the relations between projects and clusters and the difference instance groups

Network parameter is kept as a single element. On one project cases, it's the network used and on multi-project cases, it is the name of the Shared VPC network that will be created on the first project and to which the rest of the projects will be linked with.

Changed default environment from 'staging' to 'prod'

Succesful regression test performed:
$ go run ~/go/src/github.com/kubernetes-sigs/kubetest2/kubetest2-gke --up --down --num-nodes=1 --project=test-project --cluster-name=testcluster --zone=us-central1-a --network=test-network

### DIFF
--- a/kubetest2-gke/deployer/common.go
+++ b/kubetest2-gke/deployer/common.go
@@ -18,9 +18,12 @@ package deployer
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/klog"
+
 	"sigs.k8s.io/kubetest2/pkg/boskos"
 )
 
@@ -41,8 +44,8 @@ func (d *deployer) initialize() error {
 			return fmt.Errorf("init failed to verify flags for up: %s", err)
 		}
 
-		if d.project == "" {
-			klog.V(1).Info("No GCP project provided, acquiring from Boskos")
+		if len(d.projects) == 0 {
+			klog.V(1).Infof("No GCP projects provided, acquiring from Boskos %d project/s", d.boskosProjectsRequested)
 
 			boskosClient, err := boskos.NewClient(d.boskosLocation)
 			if err != nil {
@@ -50,20 +53,33 @@ func (d *deployer) initialize() error {
 			}
 			d.boskos = boskosClient
 
-			resource, err := boskos.Acquire(
-				d.boskos,
-				gkeProjectResourceType,
-				time.Duration(d.boskosAcquireTimeoutSeconds)*time.Second,
-				d.boskosHeartbeatClose,
-			)
+			for i := 0; i < d.boskosProjectsRequested; i++ {
+				resource, err := boskos.Acquire(
+					d.boskos,
+					gkeProjectResourceType,
+					time.Duration(d.boskosAcquireTimeoutSeconds)*time.Second,
+					d.boskosHeartbeatClose,
+				)
 
-			if err != nil {
-				return fmt.Errorf("init failed to get project from boskos: %s", err)
+				if err != nil {
+					return fmt.Errorf("init failed to get project from boskos: %s", err)
+				}
+				d.projects = append(d.projects, resource.Name)
+				klog.V(1).Infof("Got project %s from boskos", resource.Name)
 			}
-			d.project = resource.Name
-			klog.V(1).Infof("Got project %s from boskos", d.project)
 		}
 
+		// Multi-cluster name adjustment
+		numProjects := len(d.projects)
+		d.projectClustersLayout = make(map[string][]string, numProjects)
+		if numProjects > 1 {
+			if err := buildProjectClustersLayout(d.projects, d.clusters, d.projectClustersLayout); err != nil {
+				return fmt.Errorf("failed to build the project clusters layout: %v", err)
+			}
+		} else {
+			// Backwards compatible construction
+			d.projectClustersLayout[d.projects[0]] = d.clusters
+		}
 	}
 
 	if d.commonOptions.ShouldDown() {
@@ -72,5 +88,24 @@ func (d *deployer) initialize() error {
 		}
 	}
 
+	return nil
+}
+
+// buildProjectClustersLayout builds the projects and real cluster names mapping based on the provided --cluster-name flag.
+func buildProjectClustersLayout(projects, clusters []string, projectClustersLayout map[string][]string) error {
+	for _, clusterName := range clusters {
+		parts := strings.Split(clusterName, ":")
+		if len(parts) != 2 {
+			return fmt.Errorf("cluster name does not follow expected format (name:projectIndex): %s", clusterName)
+		}
+		projectIndex, err := strconv.Atoi(parts[1])
+		if err != nil {
+			return fmt.Errorf("cluster name does not follow contain a valid project index (name:projectIndex. E.g: cluster:0): %v", err)
+		}
+		if projectIndex >= len(projects) {
+			return fmt.Errorf("project index %d specified in the cluster name should be smaller than the number of projects %d", projectIndex, len(projects))
+		}
+		projectClustersLayout[projects[projectIndex]] = append(projectClustersLayout[projects[projectIndex]], parts[0])
+	}
 	return nil
 }
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -18,10 +18,10 @@ limitations under the License.
 package deployer
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"os"
 	realexec "os/exec" // Only for ExitError; Use kubetest2/pkg/exec to actually exec stuff
 	"path/filepath"
@@ -32,8 +32,10 @@ import (
 	"sync"
 
 	"github.com/spf13/pflag"
+	"golang.org/x/sync/errgroup"
 	"k8s.io/klog"
 	"sigs.k8s.io/boskos/client"
+
 	"sigs.k8s.io/kubetest2/pkg/build"
 	"sigs.k8s.io/kubetest2/pkg/exec"
 	"sigs.k8s.io/kubetest2/pkg/metadata"
@@ -82,20 +84,24 @@ type deployer struct {
 	// doInit helps to make sure the initialization is performed only once
 	doInit sync.Once
 	// gke specific details
-	project           string
-	zone              string
-	region            string
-	cluster           string
-	nodes             int
-	machineType       string
-	network           string
-	environment       string
-	createCommandFlag string
-	gcpServiceAccount string
+	projects []string
+	zone     string
+	region   string
+	clusters []string
+	// only used for multi-project multi-cluster profile to save the project-clusters mapping
+	projectClustersLayout map[string][]string
+	nodes                 int
+	machineType           string
+	network               string
+	subnetworkRanges      []string
+	environment           string
+	createCommandFlag     string
+	gcpServiceAccount     string
 
-	kubecfgPath    string
-	testPrepared   bool
-	instanceGroups []*ig
+	kubecfgPath  string
+	testPrepared bool
+	// project -> cluster -> instance groups
+	instanceGroups map[string]map[string][]*ig
 
 	stageLocation string
 
@@ -104,6 +110,8 @@ type deployer struct {
 
 	boskosLocation              string
 	boskosAcquireTimeoutSeconds int
+	// number of boskos projects to request if `projects` is empty
+	boskosProjectsRequested int
 
 	// boskos struct field will be non-nil when the deployer is
 	// using boskos to acquire a GCP project
@@ -133,7 +141,10 @@ func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 
 // verifyCommonFlags validates flags for up phase.
 func (d *deployer) verifyUpFlags() error {
-	if d.cluster == "" {
+	if err := d.verifyNetworkFlags(); err != nil {
+		return err
+	}
+	if d.clusters == nil {
 		return fmt.Errorf("--cluster-name must be set for GKE deployment")
 	}
 	if _, err := d.location(); err != nil {
@@ -147,10 +158,10 @@ func (d *deployer) verifyUpFlags() error {
 
 // verifyDownFlags validates flags for down phase.
 func (d *deployer) verifyDownFlags() error {
-	if d.cluster == "" {
+	if d.clusters == nil {
 		return fmt.Errorf("--cluster-name must be set for GKE deployment")
 	}
-	if d.project == "" {
+	if d.projects == nil {
 		return fmt.Errorf("--project must be set for GKE deployment")
 	}
 	if _, err := d.location(); err != nil {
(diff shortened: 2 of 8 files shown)


kubetest2-gke/deployer/common.go
```diff
@@ -74,3 +95,22 @@ func (d *deployer) initialize() error {
 
 	return nil
 }
+
+func filterMultiProjectClusterName(clusters []string, projectID int) ([]string, error) {
+	result := make([]string, 0)
+	for _, clusterName := range clusters {
+		parts := strings.Split(clusterName, "-")
```
amwat, line 102: This is a breaking change. The separator should be something that the gke cluster name doesn't allow by default.

kubetest2-gke/deployer/deployer.go
```diff
@@ -218,61 +227,156 @@ func (d *deployer) Build() error {
...
+		return err
 	}
+
+	if err := d.createNetwork(d.projects); err != nil {
+		return err
+	}
+
+	for _, project := range d.projects {
+		// TODO Prepare should be performed per GCP project
+		if err := d.prepareGcpIfNeeded(project); err != nil {
```
amwat, line 327: we shouldn't naively do this now.
prepare used to 
set up the environment (which should be a one time thing) 
and set the default project (which we shouldn't keep doing before every command) 
instead we should now make sure that all commands specify a `--project` and stop relying on setting a default project now that we have multi project.

kubetest2-gke/deployer/deployer.go
```diff
@@ -354,11 +473,7 @@ func (d *deployer) TestSetup() error {
 // Kubeconfig returns a path to a kubeconfig file for the cluster in
 // a temp directory, creating one if one does not exist.
 // It also sets the KUBECONFIG environment variable appropriately.
-func (d *deployer) Kubeconfig() (string, error) {
-	if err := d.prepareGcpIfNeeded(); err != nil {
-		return "", err
-	}
-
+func (d *deployer) Kubeconfig(project, cluster string) (string, error) {
```
amwat, line 476: This is breaking the deployerwithkubeconfig interface.
The kubeconfig returned here should be a comma separated list of all the clusters created.

kubetest2-gke/deployer/deployer.go
```diff
@@ -218,61 +227,156 @@ func (d *deployer) Build() error {
...
+		if err := d.prepareGcpIfNeeded(project); err != nil {
+			return err
+		}
+
+		loc, err := d.location()
+		if err != nil {
+			return err
+		}
+
+		for _, cluster := range d.clustersLayout[project] {
```
amwat, line 336: multi-cluster creation should definitely be parallelized!

amwat:
also there's another dupe PR https://github.com/kubernetes-sigs/kubetest2/pull/12 can we close either one of these?

kubetest2-gke/deployer/common.go
chizhg: This function will only be called `if numProjects > 1`, for single project, it will still be using the full name as the cluster name, so it's backward compatible.

kubetest2-gke/deployer/deployer.go
chizhg: It looks `kubetest2-gke` does not implement the `DeployerWithKubeconfig` interface, which defines `Kubeconfig() (string, error)` instead of `Kubeconfig(project, cluster string) (string, error)`?

Ref: https://github.com/kubernetes-sigs/kubetest2/blob/master/pkg/types/types.go#L86

chizhg:
> also there's another dupe PR #12 can we close either one of these?

Yeah we can close that one.
This PR is originally implemented by Alberto but he didn't sign the CLA and he's OOO this week, so I just copied all his change and created a new one.

kubetest2-gke/deployer/common.go
amwat: we should allow multi-cluster names to be any possible valid gke cluster name. 
otoh the choice of separator for the project suffix can be anything else.

kubetest2-gke/deployer/common.go
```diff
@@ -74,3 +95,22 @@ func (d *deployer) initialize() error {
...
+			return nil, fmt.Errorf("cluster name does not follow expected format (name-projectIndex): %s", clusterName)
+		}
+		projectIndex, err := strconv.Atoi(parts[1])
+		if err != nil {
+			return nil, fmt.Errorf("cluster name does not follow contain a valid project index (name-projectIndex. E.g: cluster-0): %s", err)
+		}
+		if projectIndex == projectID {
+			// Network field is empty
+			result = append(result, parts[0])
+		}
```
amwat, line 113: insteading of erroring if the syntax is not provided, a fallback where a random project is chosen from the provided list sounds more flexible?

kubetest2-gke/deployer/deployer.go
amwat: yeah it was previously, the current PR changes that behavior.
Also correction they should be `:` separated not comma separated
https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#linux

kubetest2-gke/deployer/common.go
chizhg: Makes sense.
Changed the separator to `:` which is not allowed in gke cluster names.

kubetest2-gke/deployer/deployer.go
chizhg: Changed to only call this function for the first GCP project, which is supposed to be the host project for the multi-project profile.

kubetest2-gke/deployer/deployer.go
chizhg: Definitely agree.
Changed to use `errgroup` to parallelize the operations.

kubetest2-gke/deployer/deployer.go
chizhg: Fixed the function signature, and used the correct separator for multi kubeconfig files - `:` for linux and MAC, `;` for windows.

kubetest2-gke/deployer/common.go
chizhg: Hmm I'm not sure if that flexibility is desirable, if users decide to use multi-project profile, it should be required to specify which project to use to create which clusters, otherwise the result won't be predictable, e.g. it could be possible to have all clusters randomly created in one project and run into quota issues. So I left this unchanged.

chizhg:
> Thanks for adding this functionality!
> 
> /ok-to-test
> 
> High level comments: can you break it down into a couple of commits that add different functionality and also move them to different files (especially for all the network stuff etc), that would make reviewing much easier.
> 
> Also since we don't have automated e2e setup for the gke deployer yet
> can you ensure that the overall flow also works?
> 
> something like `kubetest2 gke --up --down --test=exec -- <some basic testing>`
> 
> First pass comments:

Sorry for making this PR so large... I have moved the new logic to `network.go`, but it would be pretty hard for me to break the change down into multiple commits, as the changes are coupled togther, e.g. if I change the `--project` flag to support multiple, I'll have to change all the places where `d.project` is used.

I have manually tested below commands and they are working fine:
- `kubetest2 gke --up --down --test=exec -- <some basic testing>`
- `kubetest2 gke --up --down --network=test-network --test=exec -- <some basic testing>`
- `kubetest2 gke --up --test=exec -- <some basic testing>`

kubetest2-gke/deployer/deployer.go
```diff
@@ -182,19 +193,25 @@ var _ types.NewDeployer = New
 func bindFlags(d *deployer) *pflag.FlagSet {
 	flags := pflag.NewFlagSet(Name, pflag.ContinueOnError)
 
-	flags.StringVar(&d.cluster, "cluster-name", "", "Cluster name. Must be set.")
+	flags.StringSliceVar(&d.clusters, "cluster-name", []string{}, "Cluster names separated by comma. Must be set. "+
+		"For multi-project profile, it should be in the format of clusterA:1,clusterB:1,clusterC:2, where the index means the index of the project.")
```
amwat, line 197: is it 0-based or 1-based?

kubetest2-gke/deployer/deployer.go
```diff
@@ -82,20 +85,23 @@ type deployer struct {
 	// doInit helps to make sure the initialization is performed only once
 	doInit sync.Once
 	// gke specific details
-	project           string
+	projects          []string
 	zone              string
 	region            string
-	cluster           string
+	clusters          []string
+	clustersLayout    map[string][]string
```
amwat, line 92: nit: maybe rename to projectClusterLayout ?
also add a comment explaining the mapping.

kubetest2-gke/deployer/deployer.go
```diff
@@ -223,56 +240,84 @@ func (d *deployer) Up() error {
...
-		}
+	if err := d.createNetwork(); err != nil {
+		return err
+	}
+	if err := d.setupNetwork(); err != nil {
+		return err
 	}
 
+	log.Printf("Environment: %v", os.Environ())
+	eg, _ := errgroup.WithContext(context.Background())
```
amwat, line 256: nice! but we will probably need to get the Contextwithcancel here to propagate to all the cluster create calls.

kubetest2-gke/deployer/deployer.go
```diff
@@ -223,56 +240,84 @@ func (d *deployer) Up() error {
...
+					loc,
+					"--machine-type="+d.machineType,
+					"--image-type="+image,
+					"--num-nodes="+strconv.Itoa(d.nodes),
+					"--network="+d.transformNetworkName(),
+				)
+				args = append(args, subNetworkArgs...)
+				args = append(args, cluster)
+				log.Printf("Gcloud command: gcloud %+v\n", args)
+				if err := runWithOutput(exec.Command("gcloud", args...)); err != nil {
```
amwat, line 281: we might need to introduce functionality that uses https://golang.org/pkg/os/exec/#CommandContext 
in our internal package and pass the contextwithcancel generated above.

kubetest2-gke/deployer/network.go
```diff
@@ -0,0 +1,331 @@
+/*
```
amwat, line 1: Thanks for moving it to a separate file.
I'm not too aware of the functionality being added here, but mostly looks like it's following: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#creating_a_network_and_two_subnets
whigh LGTM

high level nit: let's break down most of these into helper functions (apart from createnetwork and setupnetwork) by removing the deployer receiver and pass whatever is required (projects/network names) as arguments.

kubetest2-gke/deployer/common.go
amwat: makes sense let's lave that for now. I was mostly worried about finding out too late that the user supplied syntax is wrong, but as long as we validate it early enough should be fine.

amwat:
Yeah that for breaking it into separate files, that definitely helped. 
FYI the individual commits need not be fully contained as long as it makes reviewing easier, i.e. simple adding a couple of methods in each commit is also sufficient.

You'll want to amend the commits though so that they provide relevant context. if not you can also chose to squash commits at the end, upto you.

kubetest2-gke/deployer/common.go
```diff
@@ -74,3 +96,22 @@ func (d *deployer) initialize() error {
...
+func filterClusterNamesForProject(clusters []string, index int) ([]string, error) {
+	result := make([]string, 0)
+	for _, clusterName := range clusters {
+		parts := strings.Split(clusterName, ":")
+		if len(parts) != 2 {
+			return nil, fmt.Errorf("cluster name does not follow expected format (name:projectIndex): %s", clusterName)
+		}
+		projectIndex, err := strconv.Atoi(parts[1])
+		if err != nil {
+			return nil, fmt.Errorf("cluster name does not follow contain a valid project index (name:projectIndex. E.g: cluster:0): %v", err)
```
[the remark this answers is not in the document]
chizhg: Added as the description of the flag:
`For multi-project profile, it should be in the format of clusterA:0,clusterB:1,clusterC:2, where the index means the index of the project.`

kubetest2-gke/deployer/common.go
```diff
@@ -41,29 +44,48 @@ func (d *deployer) initialize() error {
...
-			klog.V(1).Infof("Got project %s from boskos", d.project)
 		}
 
+		// Multi-cluster name adjustment
+		numProjects := len(d.projects)
+		d.clustersLayout = make(map[string][]string, numProjects)
+		if numProjects > 1 {
+			for i, project := range d.projects {
+				var err error
+				clusters, err := filterClusterNamesForProject(d.clusters, i)
```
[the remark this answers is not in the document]
chizhg: Good point!
I have changed the function to be `buildProjectClustersLayout` which will only iterate the cluster_names once.

kubetest2-gke/deployer/network.go
```diff
@@ -0,0 +1,331 @@
...
+	if len(d.projects) > 1 {
+		subnetMode = "custom"
+	}
+	// Create network if it doesn't exist.
+	// Verify networks for all projects and configure the mesh if parameter is set
+	if runWithNoOutput(exec.Command("gcloud", "compute", "networks", "describe", d.network,
+		"--project="+d.projects[0],
+		"--format=value(name)")) != nil {
+		// Assume error implies non-existent.
+		log.Printf("Couldn't describe network %q, assuming it doesn't exist and creating it", d.network)
```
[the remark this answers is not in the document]
chizhg: Done. Changed all the `log.Printf` to `klog.V(1).Infof`.

kubetest2-gke/deployer/network.go
```diff
@@ -0,0 +1,331 @@
...
+	subnetMode := "auto"
+	if len(d.projects) > 1 {
+		subnetMode = "custom"
+	}
+	// Create network if it doesn't exist.
+	// Verify networks for all projects and configure the mesh if parameter is set
+	if runWithNoOutput(exec.Command("gcloud", "compute", "networks", "describe", d.network,
+		"--project="+d.projects[0],
+		"--format=value(name)")) != nil {
+		// Assume error implies non-existent.
```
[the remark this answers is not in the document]
chizhg: This is the original code, it looks we can check that from the error message, but I would prefer not touching it in this PR.
I have added a TODO and will make the change as followups.

kubetest2-gke/deployer/network.go
```diff
@@ -0,0 +1,331 @@
...
+	// Create subnetworks for shared VPC if it's a multi-project profile.
+	// Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#creating_a_network_and_two_subnets
+	if len(d.projects) == 1 {
+		return nil
+	}
+	hostProject := d.projects[0]
+	for i, nr := range d.subnetworkRanges {
+		serviceProject := d.projects[i+1]
+		parts := strings.Split(nr, " ")
+		subnetName := d.network + "-" + serviceProject
```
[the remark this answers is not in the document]
chizhg: Done. Added the comment `The subnetwork name is in the format of `[main_network]-[service_project_id]`.`

kubetest2-gke/deployer/deployer.go
chizhg: It's 0-based, I have changed the example to have a project index as 0 to make it less confusing.

kubetest2-gke/deployer/deployer.go
chizhg: Done.

kubetest2-gke/deployer/deployer.go
chizhg: Done. I have changed to use `context.WithCancel` and call the `cancel` method whenever one cluster creation process fails.

kubetest2-gke/deployer/deployer.go
chizhg: Done. Made the change together with the above one.

kubetest2-gke/deployer/network.go
chizhg: Yes, it's mostly following that doc.
I have updated most of them into helper functions, as you suggested.

kubetest2-gke/deployer/network.go
```diff
@@ -0,0 +1,331 @@
...
+
+func (d *deployer) createNetwork() error {
+	subnetMode := "auto"
+	if len(d.projects) > 1 {
+		subnetMode = "custom"
+	}
+	// Create network if it doesn't exist.
+	// Verify networks for all projects and configure the mesh if parameter is set
+	if runWithNoOutput(exec.Command("gcloud", "compute", "networks", "describe", d.network,
+		"--project="+d.projects[0],
```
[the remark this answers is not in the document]
chizhg: Done, added the comment for both single-project and multi-project.

chizhg:
> Thanks for addressing all the comments!
> Super excited to test multi-project and multi-cluster out.
> 
> /lgtm
> /approve

Thanks a lot for reviewing! 😄

PR #15, opened 2020-08-04
Ginkgo tester automatically acquires the latest test package

michaelmdresser:
The ginkgo tester needs access to a testing package that contains test specs. This introduces a default behavior for acquiring the test package automatically from published k/k release tars.

This is an important first step to get the ginkgo tester running in Prow. We can't rely on the e2e.test package existing on the filesystem after the deployer build step because repos like kubernetes/cloud-provider-gcp don't build an e2e.test package. Future changes will add behavior to detect an already-built e2e.test package.

Tested locally.

### DIFF
--- a/pkg/testers/ginkgo/ginkgo.go
+++ b/pkg/testers/ginkgo/ginkgo.go
@@ -17,13 +17,16 @@ limitations under the License.
 package ginkgo
 
 import (
+	"archive/tar"
+	"compress/gzip"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
 	"runtime"
 	"strconv"
-	"strings"
 
 	"github.com/octago/sflags/gen/gpflag"
 	"k8s.io/klog"
@@ -33,6 +36,11 @@ import (
 
 const (
 	binary = "ginkgo" // TODO(RonWeber): Actually find these binaries.
+)
+
+var (
+	// This path is set up by AcquireTestPackage()
+	e2eTestPath = filepath.Join(os.Getenv("ARTIFACTS"), "e2e.test")
 )
 
 type Tester struct {
@@ -49,9 +57,6 @@ func (t *Tester) Test() error {
 	if err := t.pretestSetup(); err != nil {
 		return err
 	}
-
-	// This path is set up by AcquireTestPackage()
-	e2eTestPath := filepath.Join(os.Getenv("ARTIFACTS"), "kubernetes", "test", "bin", "e2e.test")
 
 	e2eTestArgs := []string{
 		"--kubeconfig=" + t.kubeconfigPath,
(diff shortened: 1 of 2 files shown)


pkg/testers/ginkgo/ginkgo.go
```diff
@@ -94,9 +96,43 @@ func (t *Tester) pretestSetup() error {
...
+		fmt.Sprintf(
+			"tar -xzf %s",
+			releaseTar),
+	}
+
+	command := "bash"
+	args := []string{
+		"-c",
+		strings.Join(script, "; "),
+	}
```
amwat, line 129: hmm do we really need a complex script for this?
seems like we can just get the name first by running `gsutil cat` and then a different command for `gsutil cp`.

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -94,9 +96,43 @@ func (t *Tester) pretestSetup() error {
...
+// is available
+func (t *Tester) AcquireTestPackage() error {
+	releaseTar := fmt.Sprintf("kubernetes-test-%s-%s.tar.gz", runtime.GOOS, runtime.GOARCH)
+
+	script := []string{
+		"set -o xtrace",
+		"set -o pipefail",
+		"set -o errexit",
+		"set -o nounset",
+		"cd $ARTIFACTS",
```
amwat, line 116: we shouldn't put this under artifacts, maybe a tempdir?

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -94,9 +96,43 @@ func (t *Tester) pretestSetup() error {
...
+		fmt.Sprintf(
+			"gsutil cp gs://kubernetes-release/release/$(gsutil cat gs://kubernetes-release/release/latest.txt)/%s .",
+			releaseTar),
+		fmt.Sprintf(
+			"tar -xzf %s",
```
amwat, line 121: can we untar by using standard golang packages ? 
https://golang.org/pkg/archive/tar/
https://golang.org/pkg/compress/gzip/

pkg/testers/ginkgo/ginkgo.go
michaelmdresser: Updated to perform commands one at a time using `exec`.

pkg/testers/ginkgo/ginkgo.go
michaelmdresser: Done. Specifically, it will now only extract the file we care about and nothing else.

pkg/testers/ginkgo/ginkgo.go
michaelmdresser: Hmm. We do put other artifacts in here, I think its a reasonable place to download to. Its the only space "owned" by kubetest2 and would let a test runner potentially inspect which version of the test package was downloaded because it will be at a known path.

michaelmdresser:
Ginkgo is acting weird.
/hold

pkg/testers/ginkgo/ginkgo.go
amwat: this will end up reuploading the entire kubernetes release tarballs in every job to gcs.
maybe the e2e.test might make sense under artifacts, but the tars should go in a tempdir which we then cleanup.

michaelmdresser: Ah, didn't realize that. Added logic to download to a tempdir and then remove the dir after extracting the test package.

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -94,9 +101,109 @@ func (t *Tester) pretestSetup() error {
...
+		return fmt.Errorf("getting latest release name had no output")
+	}
+
+	releaseName := lines[0]
+
+	// next, download the matching release tar
+
+	cmd = exec.Command("gsutil", "cp",
+		fmt.Sprintf(
+			"gs://kubernetes-release/release/%s/%s",
```
amwat, line 146: nit: extract the gcs bucket to a flag (defaulting to `kubernetes-release`) for reusability.

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -94,9 +101,109 @@ func (t *Tester) pretestSetup() error {
...
 
+	if err := t.AcquireTestPackage(); err != nil {
+		return fmt.Errorf("failed to get ginkgo test package from published releases: %s", err)
+	}
+
 	return nil
 }
 
+// TODO(michaelmdresser): change behavior if a local built e2e.test package
+// is available
```
amwat, line 112: nit: also add functionality download a specific version instead of just latest.

michaelmdresser: Done.

pkg/testers/ginkgo/ginkgo.go
michaelmdresser: Do you mean let the user optionally specify a non-temp dir to download the tar to? It's currently deleted after extracting the test package and log messages include the version if the user wants to re-acquire it manually.

amwat: no I meant the GCS bucket that's being used instead of a fixed one `gs://kubernetes-release` we should make that a flag.

michaelmdresser: Done.

PR #22, opened 2020-08-06
Automatically generate command help docs

chizhg:


chizhg:
This approach is not desirable after having some offline conversations with @amwat , and there is a better way for providing the docs coming soon. So close this PR for now.

PR #26, opened 2020-08-07
Add flags to GCE deployer to set cache mutation detector, runtime config, create custom network, and enable pod security policy

michaelmdresser:
These flags are part of the necessary features to support a drop-in replacement (using kubetest2) of the merge-blocking job pull-kubernetes-e2e-gce. See https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L37 for the original job spec.

### DIFF
--- a/kubetest2-gce/deployer/deployer.go
+++ b/kubetest2-gce/deployer/deployer.go
@@ -65,6 +65,8 @@ type deployer struct {
 	BoskosLocation              string `desc:"If set, manually specifies the location of the boskos server. If unset and boskos is needed, defaults to http://boskos.test-pods.svc.cluster.local."`
 	LegacyMode                  bool   `desc:"Set if the provided repo root is the kubernetes/kubernetes repo and not kubernetes/cloud-provider-gcp."`
 	NumNodes                    int    `desc:"The number of nodes in the cluster."`
+	EnableCacheMutationDetector bool   `desc:"Sets the environment variable ENABLE_CACHE_MUTATION_DETECTOR=true during deployment."`
+	RuntimeConfig               string `desc:"Sets the KUBE_RUNTIME_CONFIG environment variable during deployment."`
 }
 
 // New implements deployer.New for gce
--- a/kubetest2-gce/deployer/deployer.go
+++ b/kubetest2-gce/deployer/deployer.go
@@ -65,8 +65,11 @@ type deployer struct {
 	BoskosLocation              string `desc:"If set, manually specifies the location of the boskos server. If unset and boskos is needed, defaults to http://boskos.test-pods.svc.cluster.local."`
 	LegacyMode                  bool   `desc:"Set if the provided repo root is the kubernetes/kubernetes repo and not kubernetes/cloud-provider-gcp."`
 	NumNodes                    int    `desc:"The number of nodes in the cluster."`
+
 	EnableCacheMutationDetector bool   `desc:"Sets the environment variable ENABLE_CACHE_MUTATION_DETECTOR=true during deployment. This should cause a panic if anything mutates a shared informer cache."`
 	RuntimeConfig               string `desc:"Sets the KUBE_RUNTIME_CONFIG environment variable during deployment."`
+	EnablePodSecurityPolicy     bool   `desc:"Sets the environment variable ENABLE_POD_SECURITY_POLICY=true during deployment."`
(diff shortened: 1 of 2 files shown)


kubetest2-gce/deployer/deployer.go
```diff
@@ -65,6 +65,8 @@ type deployer struct {
 	BoskosLocation              string `desc:"If set, manually specifies the location of the boskos server. If unset and boskos is needed, defaults to http://boskos.test-pods.svc.cluster.local."`
 	LegacyMode                  bool   `desc:"Set if the provided repo root is the kubernetes/kubernetes repo and not kubernetes/cloud-provider-gcp."`
 	NumNodes                    int    `desc:"The number of nodes in the cluster."`
+	EnableCacheMutationDetector bool   `desc:"Sets the environment variable ENABLE_CACHE_MUTATION_DETECTOR=true during deployment."`
+	RuntimeConfig               string `desc:"Sets the KUBE_RUNTIME_CONFIG environment variable during deployment."`
```
amwat, line 69: Is this used anywhere? we should not port all the flags over from kubetest. only the ones we need.

kubetest2-gce/deployer/deployer.go
```diff
@@ -65,6 +65,8 @@ type deployer struct {
 	BoskosLocation              string `desc:"If set, manually specifies the location of the boskos server. If unset and boskos is needed, defaults to http://boskos.test-pods.svc.cluster.local."`
 	LegacyMode                  bool   `desc:"Set if the provided repo root is the kubernetes/kubernetes repo and not kubernetes/cloud-provider-gcp."`
 	NumNodes                    int    `desc:"The number of nodes in the cluster."`
+	EnableCacheMutationDetector bool   `desc:"Sets the environment variable ENABLE_CACHE_MUTATION_DETECTOR=true during deployment."`
```
amwat, line 68: can we add some basic info about what it does. 
There's a description here:
https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-cloud-provider/gcp/gcp-gce.yaml#L52

kubetest2-gce/deployer/deployer.go
michaelmdresser: The linked test in the PR description uses both of these.

kubetest2-gce/deployer/deployer.go
michaelmdresser: Added an equivalent description.

michaelmdresser:
Added the flags for the extra envs in the presets. Checked against k/k and found no mention of `KUBE_GCS_UPDATE_LATEST` (the other env in the presets) so I'm pretty sure it is prow-specific and shouldn't cause any problems.

kubetest2-gce/deployer/deployer.go
amwat: ahh, it's a separate flag not a --env 👍

kubetest2-gce/deployer/deployer.go
```diff
@@ -65,6 +65,11 @@ type deployer struct {
 	BoskosLocation              string `desc:"If set, manually specifies the location of the boskos server. If unset and boskos is needed, defaults to http://boskos.test-pods.svc.cluster.local."`
 	LegacyMode                  bool   `desc:"Set if the provided repo root is the kubernetes/kubernetes repo and not kubernetes/cloud-provider-gcp."`
 	NumNodes                    int    `desc:"The number of nodes in the cluster."`
+
+	EnableCacheMutationDetector bool   `desc:"Sets the environment variable ENABLE_CACHE_MUTATION_DETECTOR=true during deployment. This should cause a panic if anything mutates a shared informer cache."`
+	RuntimeConfig               string `desc:"Sets the KUBE_RUNTIME_CONFIG environment variable during deployment."`
+	EnablePodSecurityPolicy     bool   `desc:"Sets the environment variable ENABLE_POD_SECURITY_POLICY=true during deployment."`
+	CreateCustomNetwork         bool   `desc:"Sets the environment variable CREATE_CUSTOM_NETWORK=true during deployment."`
```
amwat, line 72: nit: in a follow up we should add atleast some description so that it's not a mystery.
we should also move these to specific structs based on what they affect.

PR #33, opened 2020-08-12
upgrade to go1.15

amwat:
/cc @BenTheElder


BenTheElder:
so blissfully easy <3

PR #34, opened 2020-08-12
GCE deployer creates firewall rule allowing nodeport access

michaelmdresser:
This issue was overlooked for a while because testing of just the deployer works fine without these firewall rules. However, once real testing began, large groups of (unexpected) failures were noticed (see https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/92316/pull-kubernetes-e2e-gce-kubetest2/1293334635354263552), especially with the entire NodePort category of tests. Adding these firewall rules (which are present [in the old test-setup bash function](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/util.sh#L3810) this deployer replaces) resolves this issue.

Tested locally. Firewalls are correctly created and deleted. Previously failing NodePort tests pass.

### DIFF
--- a/kubetest2-gce/deployer/common.go
+++ b/kubetest2-gce/deployer/common.go
@@ -128,12 +128,16 @@ func (d *deployer) buildEnv() []string {
 
 	// kube-up and kube-down get this as a default ("kubernetes") but log-dump
 	// does not. opted to set it manually here for maximum consistency
-	env = append(env, "KUBE_GCE_INSTANCE_PREFIX=kubetest2")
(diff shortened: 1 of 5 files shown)


amwat:
FYI @cheftako @cici37

This brings the failures down to just 3: https://testgrid.k8s.io/provider-gcp-periodics#Conformance%20-%20Cloud%20Provider%20GCP%20-%20master

cheftako:
Awesome! Great work!

PR #36, opened 2020-08-12
Fix a few issues for the kubetest2-gke deployer and add chizhg as reviewer

chizhg:
1. Print a warning message instead of returning an error for any VPC errors, since sometimes we may want to use the projects pre-configured with shared-vpc for testing, then the service account that runs this command doesn't necessarily need to have the permissions

2. Add a new `--enable-workload-identity` flag which can be used to enable workload identity for the GKE clusters

3. Add myself as the reviewer of kubetest2 gke deployer

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/commandutils.go
+++ b/kubetest2-gke/deployer/commandutils.go
@@ -63,7 +63,7 @@ func (d *deployer) prepareGcpIfNeeded(projectID string) error {
 		return err
 	}
 
-	if d.gcpSSHKeyRequired {
+	if d.gcpSSHKeyIgnored {
 		// Ensure ssh keys exist
 		klog.V(1).Info("Checking existing of GCP ssh keys...")
 		k := filepath.Join(home(".ssh"), "google_compute_engine")
--- a/kubetest2-gke/OWNERS
+++ b/kubetest2-gke/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+reviewers:
+- chizhg
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -109,7 +109,10 @@ type deployer struct {
 	gcsLogsDir   string
 
 	// whether the GCP SSH key is required or not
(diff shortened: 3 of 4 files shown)


kubetest2-gke/deployer/commandutils.go
```diff
@@ -55,7 +55,13 @@ func (d *deployer) prepareGcpIfNeeded(projectID string) error {
 	}
 
 	if err := runWithOutput(exec.Command("gcloud", "config", "set", "project", projectID)); err != nil {
-		return fmt.Errorf("Failed to set project %s : err %v", projectID, err)
+		return fmt.Errorf("failed to set project %s : err %v", projectID, err)
+	}
+
+	// TODO(chizhg):Remove this once we add --enable-workload-identity as the deployer flag,
+	//              then the PROJECT won't need to be specified in the cluster creation command.
```
amwat, line 62: hmm i would prefer to avoid porting over existing kubetest behavior.
Can we directly add the workload identity option?

chizhg: The main reason I'm doing this way is adding the workload identity option will involve changes in 3 different places 😢 
But you are right, that would be way cleaner and this request is not that urgent, so I will add the option in this PR.

chizhg: Added the workload identity option.

chizhg:
> FYI you can run `make verify` locally.
> 
> I am willing to unblock existing scripts, but i would discourage using custom create commands in general.
> That being said, adding the workload identity flag directly seems straightforward enough?

It looks using custom create commands is usually unavoidable, for the create command we use in Knative:
```
container clusters create --quiet --enable-autoscaling
      --min-nodes=${E2E_MIN_CLUSTER_NODES} --max-nodes=${E2E_MAX_CLUSTER_NODES}
      --cluster-version=${E2E_CLUSTER_VERSION}
      --scopes=cloud-platform --enable-basic-auth --no-issue-client-certificate
      --no-enable-ip-alias --no-enable-autoupgrade
     ${GKE_ADDONS}
```

At least `--cluster-version`, `--scopes`, `--enable-basic-auth` are required.

PR #40, opened 2020-08-14
Enable build caching in ginkgo tester

amwat:
Checks if there is an existing downloaded release tar and verifies it's hash.
Add a new --test-args flag to enable passing additional flags to the e2e.test binary. 
Use klog instead of log.
Fix a unit test failing for a common use case on MacOS where `/tmp -> /private/tmp`.

/cc @michaelmdresser

### DIFF
--- a/pkg/testers/ginkgo/package.go
+++ b/pkg/testers/ginkgo/package.go
@@ -1,0 +1,200 @@
+package ginkgo
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"k8s.io/klog"
+	"sigs.k8s.io/kubetest2/pkg/exec"
+)
+
+// AcquireTestPackage obtains two test binaries and places them in $ARTIFACTS.
+// The first is "ginkgo", the actual ginkgo executable.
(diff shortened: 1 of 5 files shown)


pkg/testers/ginkgo/package.go
```diff
@@ -0,0 +1,200 @@
...
+// and verifies the hashes
+// else downloads it from GCS
+func (t *Tester) ensureReleaseTar(downloadPath, releaseTar string) error {
+	if _, err := os.Stat(downloadPath); err == nil {
+		klog.V(0).Infof("Found existing tar at %v", downloadPath)
+		if err := t.compareSHA(downloadPath, releaseTar); err == nil {
+			klog.V(0).Infof("Validated hash for existing tar at %v", downloadPath)
+			return nil
+		}
+		klog.Warning(err)
```
michaelmdresser, line 145: I think it would be better to explicitly check if the err with `os.IsNotExist(err)` and log a non-warning message if the problem is simply that it needs to be downloaded. If it is a different error, then this should fail immediately.

amwat: I think in either case we want to proceed with downloading? which will overwrite if hashes don't match etc.

michaelmdresser: True. Just a little worried about a different error suggesting some filesystem issue, but I imagine we would have caught it already? Should be fine.

PR #44, opened 2020-08-14
Some fixes to kubetest2 GKE deployer

chizhg:
Fix a few subtle bugs for kubetest2 GKE deployer, mainly for firewall rules and networking for multicluster profile.

I have done the manual regression tests for both single cluster and multi-cluster, and nothing breaks.

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/firewall.go
+++ b/kubetest2-gke/deployer/firewall.go
@@ -20,22 +20,21 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"k8s.io/klog"
 
 	"sigs.k8s.io/kubetest2/pkg/exec"
 )
 
-func (d *deployer) ensureFirewall(project, cluster, network string) error {
+func (d *deployer) ensureFirewall(hostProject, curtProject, cluster, network string) error {
+	klog.V(1).Infof("Ensuring firewall rules for cluster %s in %s", cluster, curtProject)
 	if network == "default" {
 		return nil
 	}
-	firewall, err := d.getClusterFirewall(project, cluster)
-	if err != nil {
-		return fmt.Errorf("error getting unique firewall: %v", err)
-	}
+	firewall := d.getClusterFirewall(curtProject, cluster)
 	if runWithNoOutput(exec.Command("gcloud", "compute", "firewall-rules", "describe", firewall,
-		"--project="+project,
+		"--project="+hostProject,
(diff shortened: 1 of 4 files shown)


kubetest2-gke/deployer/firewall.go
```diff
@@ -91,17 +93,20 @@ func (d *deployer) cleanupNetworkFirewalls(project, network string) (int, error)
...
-		commandArgs = append(commandArgs, "--project="+project)
+		commandArgs = append(commandArgs, "--project="+hostProject)
 		errFirewall := runWithOutput(exec.Command("gcloud", commandArgs...))
 		if errFirewall != nil {
 			return 0, fmt.Errorf("error deleting firewall: %v", errFirewall)
 		}
-		return len(fwList), nil
+		// Sleep 10 seconds to wait for the firewall rules being deleted completely.
+		// TODO(chizhg): change to a more reliable way to check if they are deleted or not.
+		time.Sleep(10 * time.Second)
```
amwat, line 103: does this mean the `gcloud firewall-rules delete` command exits before the firewalls are actually deleted?

chizhg: Based on my experiment, yes, sometimes I got "network is still used by firewall rule xxx", even if the log shows the firewall rule has been deleted.
I have changed the comment to include this information.

chizhg:
> The pre-existing functionality seems to rely too much on top level struct fields 😕
> I'll most likely refactor them into separate options in the future.
> 
> /lgtm
> /approve
> /hold
> for one question

Do you mean too many functions are using `deployer` as the receiver? If yes definitely agree...

amwat:
/lgtm
/hold cancel

yeah because everything is in the top level struct. 

They should be split into modular options e.g. networkOptions etc.

mrfaizal:
any thoughts on adding unit tests?

chizhg:
> any thoughts on adding unit tests?

There is little to no unit test in this repo now :-(

And unfortunately the current way we implement `kubetest2-gke` makes it almost impossible to unit test, since it's using `gcloud` command under-the-hood. https://github.com/knative/test-infra/tree/master/kntest/pkg/cluster/gke is a similar tool implemented for Knative a while ago, which is using the Google Cloud Go APIs and has pretty high test coverage. I had some prior discussions with Amit and Ben on moving it here, and will likely do it in the next quarter, then we can have more tests added.

mrfaizal:
sgtm!

PR #46, opened 2020-08-28
gke deployer supports creating private cluster

chizhg:
gke deployer supports creating private cluster, by following https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters

/cc @amwat


kubetest2-gke/deployer/up.go
```diff
@@ -78,6 +79,7 @@ func (d *deployer) Up() error {
 					args = append(args, fmt.Sprintf("--workload-pool=%s.svc.id.goog", project))
 				}
 				args = append(args, subNetworkArgs...)
+				args = append(args, privateClusterArgs...)
```
amwat, line 82: +1 this is the pattern i'd want in the future too!
i.e. each set of options will return the args that we pass to the create call

PR #50, opened 2020-09-10
Use gcloud beta command group for shared vpc operations

chizhg:
It turns out shared VPC is still in beta, so we'll need to use gcloud beta command group for shared VPC operations.
(The doc in https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#enabling_and_granting_roles does not reflect it though..)

/cc @amwat


amwat (commented):
https://cloud.google.com/vpc/docs/provisioning-shared-vpc#enable-shared-vpc-host

/lgtm
/approve

PR #53, opened 2020-09-15
[Clusterloader2] Implement a clusterloader2 tester

amwat:
Create an initial implementation of clusterloader2 tester.
Currently requires having https://github.com/kubernetes/perf-tests checked out.

/cc @BenTheElder

### DIFF
--- a/kubetest2-kind/deployer/deployer.go
+++ b/kubetest2-kind/deployer/deployer.go
@@ -63,6 +63,17 @@ type deployer struct {
 	verbosity      int    // --verbosity for kind
 }
 
+func (d *deployer) Kubeconfig() (string, error) {
+	if d.kubeconfigPath != "" {
+		return d.kubeconfigPath, nil
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".kube", "config"), nil
+}
+
 // helper used to create & bind a flagset to the deployer
 func bindFlags(d *deployer) *pflag.FlagSet {
 	flags := pflag.NewFlagSet(Name, pflag.ContinueOnError)
@@ -90,8 +101,8 @@ func bindFlags(d *deployer) *pflag.FlagSet {
 	return flags
 }
 
-// assert that deployer implements types.Deployer
-var _ types.Deployer = &deployer{}
+// assert that deployer implements types.DeployerWithKubeconfig
+var _ types.DeployerWithKubeconfig = &deployer{}
 
 // Deployer implementation methods below
 
--- a/pkg/testers/clusterloader2/cl2.go
+++ b/pkg/testers/clusterloader2/cl2.go
@@ -21,7 +21,7 @@ type Tester struct {
 	Provider      string `desc:"The type of cluster provider used (e.g gke, gce, skeleton)"`
 	KubeConfig    string `desc:"Path to kubeconfig"`
 	RepoRoot      string `desc:"Path to repository root of kubernetes/perf-tests"`
-	Nodes int `desc:"Number of nodes in the cluster. 0 will auto-detect schedulable nodes."`
+	Nodes         int    `desc:"Number of nodes in the cluster. 0 will auto-detect schedulable nodes."`
 }
 
 func NewDefaultTester() *Tester {
--- a/pkg/testers/clusterloader2/cl2.go
+++ b/pkg/testers/clusterloader2/cl2.go
@@ -1,0 +1,114 @@
+package clusterloader2
+
+import (
+	"flag"
+	"fmt"
+	"os"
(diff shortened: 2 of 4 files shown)


kubetest2-kind/deployer/deployer.go
```diff
@@ -63,6 +63,17 @@ type deployer struct {
 	verbosity      int    // --verbosity for kind
 }
 
+func (d *deployer) Kubeconfig() (string, error) {
```
BenTheElder, line 66: this isn't quite correct vs kind / client-go FYI ... https://github.com/kubernetes-sigs/kind/tree/master/pkg/cluster/internal/kubeconfig/internal/kubeconfig

pkg/testers/clusterloader2/cl2.go
```diff
@@ -0,0 +1,114 @@
...
+	"sigs.k8s.io/kubetest2/pkg/exec"
+	suite "sigs.k8s.io/kubetest2/pkg/testers/clusterloader2/suite"
+)
+
+type Tester struct {
+	Suites        string `desc:"Comma separated list of standard scale testing suites e.g. load, density"`
+	TestOverrides string `desc:"Comma separated list of paths to the config override files. The latter overrides take precedence over changes in former files."`
+	TestConfigs   string `desc:"Comma separated list of paths to test config files."`
+	Provider      string `desc:"The type of cluster provider used (e.g gke, gce, skeleton)"`
+	KubeConfig    string `desc:"Path to kubeconfig"`
```
BenTheElder, line 22: shouldn't we inject the kubeconfig env to all testers?

BenTheElder: (this also allows us to do multi-file for multi-cluster)

kubetest2-kind/deployer/deployer.go
```diff
@@ -90,8 +101,8 @@ func bindFlags(d *deployer) *pflag.FlagSet {
 	return flags
 }
 
-// assert that deployer implements types.Deployer
-var _ types.Deployer = &deployer{}
+// assert that deployer implements types.DeployerWithKubeconfig
```
BenTheElder, line 104: Is there any deployer that doesn't have kubeconfig?

pkg/testers/clusterloader2/cl2.go
amwat: we already do, this is an override. And yeah clusterloader2 doesn't support multi kubeconfig yet :/

kubetest2-kind/deployer/deployer.go
amwat: Don't think so

kubetest2-kind/deployer/deployer.go
amwat: are you referring to https://github.com/kubernetes-sigs/kind/blob/master/pkg/cluster/internal/kubeconfig/internal/kubeconfig/paths.go#L30-L43 ?

I think we are only missing the KUBECONFIG env handling from there. will add as a follow up. This should solve the basic case when nothing is set.

kubetest2-kind/deployer/deployer.go
BenTheElder: I think we should make this a standard part of the API for deployer, a requirement for kubetest2, but that can be a follow-up.

kubetest2-kind/deployer/deployer.go
BenTheElder: ~HOME is also more complex on windows, off the top of my head.

I think this deployer should probably have it's own stable computed paths and use those consistently (like $HOME/.kube/kubetest2-kind-$cluster)

pkg/testers/clusterloader2/cl2.go
BenTheElder: the usage text should probably clarify that this is an override?

BenTheElder:
/lgtm
/approve
/hold
some nits

kubetest2-kind/deployer/deployer.go
amwat: ack

pkg/testers/clusterloader2/cl2.go
amwat: done

kubetest2-kind/deployer/deployer.go
amwat: yeah, I'm planning on sending kind deployer specific changes separately soon and the top-level artifacts organization changes, will fix this more robustly there.

Current change is only to make quick testing of the clusterloader2 tester locally with the kind deployer easy.

amwat:
/hold cancel 
most of the actual kind deployer specific changes will be done in a follow up.
This PR will now enable basic clusterloader2 testing.

PR #52, opened 2020-09-14
[GKE deployer] Enable building from source 

amwat:
Add build options to control building kubernetes from source.
Refactor and extract out common functionality to pkg/
Start consolidating flags into individual options.

/cc @BenTheElder

PR #54, opened 2020-09-17
[GKE] Less strict version checking

amwat:
Also fixed a bug in down when no projects are acquired in up.

### DIFF
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -198,7 +198,7 @@ func validateVersion(version string) error {
 	case "latest":
 		return nil
 	default:
-		re, err := regexp.Compile(`(\d)\.(\d)+\.(\d)*(-gke\.\d*\.\d*)(.*)`)
+		re, err := regexp.Compile(`(\d)\.(\d)+\.(\d)*(.*)`)
 		if err != nil {
 			return err
 		}
--- a/kubetest2-gke/deployer/down.go
+++ b/kubetest2-gke/deployer/down.go
@@ -28,49 +28,49 @@ func (d *deployer) Down() error {
 	if err := d.init(); err != nil {
 		return err
 	}
+	if len(d.projects) > 0 {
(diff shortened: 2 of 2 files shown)


chizhg:
This change broke one of our test flow - https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-serving-contour-latest-beta-prow-tests/1306755852929077248

```
2020/09/18 00:48:27 Running kubetest2 with flags: ["gke" "--ignore-gcp-ssh-key=true" "--up" "--create-command=beta container clusters create --quiet --enable-autoscaling --min-nodes=4 --max-nodes=4 --scopes=cloud-platform --cluster-version=latest --addons=NodeLocalDNS" "--cluster-name=e2e-cls" "--environment=prod" "--num-nodes=4" "--machine-type=e2-standard-4" "--network=e2e-network" "--boskos-acquire-timeout-seconds=1200" "--region=us-central1" "--test=exec" "--" "/home/prow/go/src/knative.dev/serving/test/e2e-tests.sh" "--run-tests" "--contour-version" "latest"]

Error: init failed to verify flags for up: unknown version ""
```

@amwat @BenTheElder could you cc me for the future kubetest2 gke deployer update? It's now on a critical path.

amwat:
+1 definitely
FWIW this change is just making it less strict and was introduced in https://github.com/kubernetes-sigs/kubetest2/pull/52/files#diff-118cf9c7179a5f4a77c1ee75f97ebb33R201

In general i would recommend pinning to a specific commit though `go get sigs.k8s.io/kubetest2@commit`

I guess the version defaulting got lost in some of the commits in the PR (https://github.com/kubernetes-sigs/kubetest2/pull/52/commits/2c78ebf7ce2a42d98f97a4a58aedf650b8311aad#diff-118cf9c7179a5f4a77c1ee75f97ebb33R155)

will send a fix.

BenTheElder:
@amwat it might be time to start tagging versions even if they are 0.0.X, so as to have clean pins & change notes

PR #57, opened 2020-09-21
Fix firewall rules for multiproject profile

chizhg:
This PR does the following:

1. Follow https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#creating_firewall_rules to create the firewall rules for shared VPC. Still use the default `e2eAllow` firewall rules for single project.

2. Clean up the code the bit to remove `deployer` as the receiver for some simple functions.

3. Make the hack sleep a bit longer for the `cleanupNetworkFirewalls` function. Since normally we do not use `--down` in CI as we use `boskos-janitor` for cleanups, this change should only impact local runs.

--- 

I have tested the changes for multi-project multi-cluster, single project multi-cluster and single cluster, and nothing breaks (I'll submit a PR to add e2e presubmit tests for gke deployer soon).

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/firewall.go
+++ b/kubetest2-gke/deployer/firewall.go
@@ -27,53 +27,88 @@ import (
 	"sigs.k8s.io/kubetest2/pkg/exec"
 )
 
-func (d *deployer) ensureFirewall(hostProject, curtProject, cluster, network string) error {
-	klog.V(1).Infof("Ensuring firewall rules for cluster %s in %s", cluster, curtProject)
-	if network == "default" {
+func (d *deployer) ensureFirewallRules() error {
+	// Do not modify the firewall rules for the default network
+	if d.network == "default" {
 		return nil
 	}
-	firewall := d.getClusterFirewall(curtProject, cluster)
-	if runWithNoOutput(exec.Command("gcloud", "compute", "firewall-rules", "describe", firewall,
-		"--project="+hostProject,
-		"--format=value(name)")) == nil {
-		// Assume that if this unique firewall exists, it's good to go.
-		return nil
-	}
-	klog.V(1).Infof("Couldn't describe firewall '%s', assuming it doesn't exist and creating it", firewall)
 
-	tagOut, err := exec.Output(exec.Command("gcloud", "compute", "instances", "list",
-		"--project="+curtProject,
-		"--filter=metadata.created-by:*"+d.instanceGroups[curtProject][cluster][0].path,
-		"--limit=1",
-		"--format=get(tags.items)"))
-	if err != nil {
-		return fmt.Errorf("instances list failed: %s", execError(err))
-	}
-	tag := strings.TrimSpace(string(tagOut))
-	if tag == "" {
-		return fmt.Errorf("instances list returned no instances (or instance has no tags)")
+	if len(d.projects) == 1 {
+		project := d.projects[0]
+		return ensureFirewallRulesForSingleProject(project, d.network, d.projectClustersLayout[project], d.instanceGroups)
 	}
 
-	if err := runWithOutput(exec.Command("gcloud", "compute", "firewall-rules", "create", firewall,
-		"--project="+hostProject,
-		"--network="+network,
-		"--allow="+e2eAllow,
-		"--target-tags="+tag)); err != nil {
-		return fmt.Errorf("error creating e2e firewall: %v", err)
+	return ensureFirewallRulesForMultiProjects(d.projects, d.network, d.subnetworkRanges)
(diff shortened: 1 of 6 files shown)


kubetest2-gke/deployer/firewall.go
```diff
@@ -99,9 +134,9 @@ func (d *deployer) cleanupNetworkFirewalls(hostProject, network string) (int, er
 			return 0, fmt.Errorf("error deleting firewall: %v", errFirewall)
 		}
 		// It looks sometimes gcloud exits before the firewall rules are actually deleted,
-		// so sleep 10 seconds to wait for the firewall rules being deleted completely.
+		// so sleep 30 seconds to wait for the firewall rules being deleted completely.
 		// TODO(chizhg): change to a more reliable way to check if they are deleted or not.
-		time.Sleep(10 * time.Second)
+		time.Sleep(30 * time.Second)
```
amwat, line 139: hmm, this is starting to become unreliable now. Do we use the firewall-rules immediately after anywhere to need to wait till we ensure it's actually deleted?

I think we should just return as soon as gcloud returns (even if it's running some hidden operation at the back that actually deletes asynchronously).

from kubetest2 perspective this is a best-effort cleanup and if gcloud says it's done deleting (by returning an exit code) then we shouldn't be waiting.

kubetest2-gke/deployer/firewall.go
```diff
@@ -27,53 +27,88 @@ import (
...
+// Ensure firewall rules for multi-project profile.
+// Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#creating_additional_firewall_rules
+// Please note we are not including the firewall rule for SSH connection as it's not needed for testing.
+func ensureFirewallRulesForMultiProjects(projects []string, network string, subnetworkRanges []string) error {
+	hostProject := projects[0]
+	for i := 1; i < len(projects); i++ {
+		curtProject := projects[i]
+		firewall := fmt.Sprintf("%s-rule-%s", hostProject, curtProject)
+		// sourceRanges need to be separated with ",", while the provided subnetworkRanges are separated with space.
+		sourceRanges := strings.ReplaceAll(subnetworkRanges[i-1], " ", ",")
```
amwat, line 98: do we need them space separated as input?
let's just get them comma separated as input too?

(we should still keep this logic for backwards compatibility)

amwat (commented):
/approve

Some minor comments:

kubetest2-gke/deployer/firewall.go
chizhg: Yeah it is pretty annoying, if I don't add the sleep (or sleep shorter like 10 sec), there will be an error in the next steps of deleting the subnetwork, the error message is like `ERROR: (gcloud.compute.networks.delete) Could not fetch resource:
 - The network resource 'projects/test-project-chizhg/global/networks/test-network5' is already being used by 'projects/test-project-chizhg/global/firewalls/test-network5-odm2l5xrgkhkgaqqdwsg4qbt'`. 

It could be a bug of `gcloud` but adding the hacky sleep here is the only easy way I can find to make it working.

kubetest2-gke/deployer/firewall.go
chizhg: We are using comma to separate the subnetwork ranges for different projects, see https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gke/deployer/deployer.go#L252, so we cannot change each range to be comma separated.

kubetest2-gke/deployer/firewall.go
```diff
@@ -27,53 +27,88 @@ import (
...
-	return "e2e-ports-" + d.instanceGroups[project][cluster][0].uniq
+	return "e2e-ports-" + instanceGroups[project][cluster][0].uniq
+}
+
+// Ensure firewall rules for multi-project profile.
+// Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#creating_additional_firewall_rules
+// Please note we are not including the firewall rule for SSH connection as it's not needed for testing.
+func ensureFirewallRulesForMultiProjects(projects []string, network string, subnetworkRanges []string) error {
+	hostProject := projects[0]
+	for i := 1; i < len(projects); i++ {
```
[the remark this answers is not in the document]
chizhg: Based on what I observed, it's pretty fast to create a new firewall rule (around 1~2 second), which is rather short compared to cluster creation time, so I think we don't need to parallelize it for now.

amwat (commented):
/lgtm

PR #59, opened 2020-09-22
[GKE] Enable Log dumping

amwat:
Move repo-root to top-level since it's also required by dumplogs.
Add automatic log dumping on cluster create.

### DIFF
--- a/kubetest2-gke/deployer/options/build.go
+++ b/kubetest2-gke/deployer/options/build.go
@@ -9,16 +9,12 @@ import (
 type BuildOptions struct {
 	Strategy      string `flag:"~strategy" desc:"Determines the build strategy to use either make or bazel."`
 	StageLocation string `flag:"~stage" desc:"Upload binaries to gs://bucket/ci/job-suffix if set"`
-	RepoRoot      string `flag:"~repo-root" desc:"Path to root of the kubernetes repo. Used with --build."`
-
+	RepoRoot      string `flag:"-"`
 	build.Builder
 	build.Stager
 }
 
 func (bo *BuildOptions) Validate() error {
-	if bo.RepoRoot == "" {
-		return fmt.Errorf("required repo-root when building from source")
-	}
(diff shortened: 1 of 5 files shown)


kubetest2-gke/deployer/up.go
```diff
@@ -38,6 +38,16 @@ func (d *deployer) Up() error {
 		return err
 	}
 
+	defer func() {
+		if d.RepoRoot == "" {
+			klog.Warningf("repo-root not supplied, skip dumping cluster logs")
```
chizhg, line 43: I feel this should be `Info` since it's normal for non-k8s projects to skip dumping clusters logs.

kubetest2-gke/deployer/options/build.go
```diff
@@ -9,16 +9,12 @@ import (
 type BuildOptions struct {
 	Strategy      string `flag:"~strategy" desc:"Determines the build strategy to use either make or bazel."`
 	StageLocation string `flag:"~stage" desc:"Upload binaries to gs://bucket/ci/job-suffix if set"`
-	RepoRoot      string `flag:"~repo-root" desc:"Path to root of the kubernetes repo. Used with --build."`
-
+	RepoRoot      string `flag:"-"`
```
chizhg, line 12: Could you explain how `-` works here? Is this change related to the other `RepoRoot` field added in the `deployer` struct?

kubetest2-gke/deployer/up.go
amwat: Usually info comments are used for more verbose explanation of steps being carried out.
Whereas here we'll be missing a key part of the deployer lifecycle, feels like it should be a warning.

kubetest2-gke/deployer/options/build.go
amwat: Yup, `-` means this field will not be exposed as a flag: https://github.com/octago/sflags#options-for-flag-tag
now that the top level field is exposed.

PR #60, opened 2020-09-23
Fix a bug in creating private cluster for gke deployer

chizhg:
The key and value for the`--create-subnetwork` flag should be either separated into two command args, or connected with equal.

/cc @amwat


amwat (commented):
/lgtm
/approve

PR #61, opened 2020-09-25
Stage sources via k/release API (krel)

saschagrunert:
We now stage the sources directly via the API provided by k/release rather than the deprecated push-build.sh script (lives in k/release, too).

Side note: The API has not been e2e tested yet. I hope that we can do this with an additional job in test-infra which uses kubetest2 to periodically stage a built release.


go.mod
```diff
@@ -3,13 +3,14 @@ module sigs.k8s.io/kubetest2
...
+	github.com/google/uuid v1.1.2
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/octago/sflags v0.2.0
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.0.0
 	github.com/spf13/pflag v1.0.5
-	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
+	golang.org/x/sync v0.0.0-20200625203802-6e8e738ad208
 	k8s.io/klog v1.0.0
+	k8s.io/release v0.4.2-0.20200925090312-5d424b99ed9c // TODO: stick to the master only for testing
```
amwat, line 14: I would want to know the full repercussions here. We've previously already made efforts to minimize the dependencies especially on k8s.io since they are notoriously hard to maintain.

/cc @spiffxp @BenTheElder

amwat (commented):
Really love that we are moving away from the push-build script!

FYI if you want to test this locally this flow gets triggered by
```
kubetest2 gke --build --strategy=make --stage=gs://your-bucket
```

go.mod
BenTheElder: this is my biggest concern. I see imports like https://github.com/kubernetes/release/blob/89fddae9b1044b99146575ad89e18ce3b929bee2/go.mod#L33 upstream, which is kind of a nasty dependency for a project intended to be embedded in other repos.

I think we probably want avoid importing staging repos and k/k in kubetest2, I'm not sure the best path forward here though.

BenTheElder: maybe we can break out the relevant common package(s) somewhere?

saschagrunert: Hm, the kubectl dependency is used by different parts of k/release, If I run `go mod vendor` then I only see a very minimal subset of k/release actually being used for this implementation:

```
> find vendor/k8s.io -type d
vendor/k8s.io
vendor/k8s.io/release
vendor/k8s.io/release/pkg
vendor/k8s.io/release/pkg/git
vendor/k8s.io/release/pkg/http
vendor/k8s.io/release/pkg/release
vendor/k8s.io/release/pkg/release/regex
vendor/k8s.io/release/pkg/command
vendor/k8s.io/release/pkg/util
vendor/k8s.io/release/pkg/gcp
vendor/k8s.io/release/pkg/gcp/gcs
vendor/k8s.io/klog
vendor/k8s.io/utils
vendor/k8s.io/utils/pointer
vendor/k8s.io/apimachinery
vendor/k8s.io/apimachinery/pkg
vendor/k8s.io/apimachinery/pkg/util
vendor/k8s.io/apimachinery/pkg/util/validation
vendor/k8s.io/apimachinery/pkg/util/validation/field
vendor/k8s.io/apimachinery/pkg/util/sets
vendor/k8s.io/apimachinery/pkg/util/errors
vendor/k8s.io/test-infra
vendor/k8s.io/test-infra/prow
vendor/k8s.io/test-infra/prow/logrusutil
vendor/k8s.io/test-infra/prow/config
vendor/k8s.io/test-infra/prow/config/secret
vendor/k8s.io/test-infra/prow/version
```

Since we would like to re-use parts of the push logic for the anago transition it would be more complicated for us to split-up the repo… OTOH we could think about having multiple `go.mod`s inside k/release.

cc @justaugustus @tpepper @puerco

BenTheElder: > Hm, the kubectl dependency is used by different parts of k/release, If I run go mod vendor then I only see a very minimal subset of k/release actually being used for this implementation:

I'd want some strong guarantees that we will not have anything like this as a transitive dependency in kubetest2, we want to let people extending this repo import k8s libraries freely without conflicting with this repo.

> Since we would like to re-use parts of the push logic for the anago transition it would be more complicated for us to split-up the repo… OTOH we could think about having multiple go.mods inside k/release.

I mean, if you split it up into a util / re-usable packages repo and the repo implementing anago etc. that should be fine, similar to k/k and k8s.io/utils (which is not staged, it's imported).

I'm not sure how well multiple modules in a repo works when you want to import one module from another.
I know it works OK for modules that do not import each other.

justaugustus: The culprit here is `k8s.io/kubectl/pkg/cmd/util/editor` which is pulled in by the new release notes editing flow.
If that could be extracted from k/k as a general purpose CLI editor, we could clean up the imports.

Thoughts @soltysh @pwittrock @seans3?

puerco: I can yank out kubectl fairly easy and with that, I think we can strip out the apimachinery dependency too.

puerco: I've removed the dependecies to both [kubectl/…/editor](kubernetes/release#1599) and the errors package in [api-machinery](kubernetes/release#1598), this problem should be gone once those merge.

saschagrunert:
This is ready for review. PTAL

Unfortunately I was not able to test it locally:
```
> go run main.go gke --build --strategy=make --stage=gs://kubernetes-release-gcb/kubetest2-test
Error: could not find kubetest2 deployer "gke"
```

amwat:
`make install` and `make install-deployer-gke` should give you the necessary binaries for local testing.

saschagrunert:
> `make install` and `make install-deployer-gke` should give you the necessary binaries for local testing.

Thanks! I think the path for the bazel release tars was wrong there:
https://github.com/kubernetes-sigs/kubetest2/blob/237f174c03bcea9eb4a838ddfea1fc41a46a7ece/pkg/build/make.go#L82-L86

Fixed it in this PR. Now it seems to work:
```
> kubetest2 gke \
    --repo-root=$PWD \
    --build \
    --strategy=make \
    --stage=gs://kubernetes-release-gcb/ci/kubetest2
…
INFO[0413] Using Bazel build version
INFO[0413] Found build version: v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0413] BuildDir is not set, setting it automatically
INFO[0413] Release is build by bazel, setting BuildDir to bazel-bin/build
INFO[0413] Latest version is v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0413] Checking bucket kubernetes-release-gcb for write permissions
INFO[0413] Staging local artifacts
INFO[0413] Cleaning staging dir bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0413] Removing bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0413] Creating bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0413] Copying release tarballs
INFO[0413] Trying to copy dir bazel-bin/build/release-tars to bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0413] Trying to copy file bazel-bin/build/release-tars/.dummy to bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a/.dummy (required: false)
INFO[0413] Copied .dummy
INFO[0413] Trying to copy file bazel-bin/build/release-tars/_client-bin.args to bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a/_client-bin.args (required: false)
INFO[0413] Copied _client-bin.args
INFO[0413] Trying to copy file bazel-bin/build/release-tars/_client-bin.tar to bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a/_client-bin.tar (required: false)
INFO[0414] Copied _client-bin.tar
…
INFO[0557] Done
INFO[0557] Skipping not existing extra dir bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a/extra
INFO[0557] Skipping not existing plain binaries dir bazel-bin/build/release-stage
INFO[0557] Writing checksums
INFO[0557] Writing artifact hashes to SHA256SUMS/SHA512SUMS files
INFO[0702] Trying to copy file SHA256SUMS to bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a/SHA256SUMS (required: true)
INFO[0702] Copied SHA256SUMS
INFO[0702] Trying to copy file SHA512SUMS to bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a/SHA512SUMS (required: true)
INFO[0702] Copied SHA512SUMS
INFO[0702] Hashing files in bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0832] GCS destination is ci/kubetest2/v1.20.0-alpha.1.227_89fdf5e7d3769a
INFO[0832] Pushing release artifacts from bazel-bin/build/gcs-stage/v1.20.0-alpha.1.227_89fdf5e7d3769a to gs://kubernetes-release-gcb/ci/kubetest2/v1.20.0-alpha.1.227_89fdf5e7d3769a
Building synchronization state...
Starting synchronization...
…
```
Artifacts are located there: https://console.cloud.google.com/storage/browser/kubernetes-release-gcb/ci?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false

Depends on some fixes to be merged in https://github.com/kubernetes/release/pull/1629/files

pkg/build/make.go
```diff
@@ -64,7 +64,7 @@ func K8sDir(topdir string, parts ...string) (string, error) {
 // TODO(RonWeber): This whole untarring and cd-ing logic is out of
 // scope for Build().  It needs a better home.
 func extractBuiltTars() error {
-	allBuilds, err := K8sDir("kubernetes", "_output", "gcs-stage")
+	allBuilds, err := K8sDir("kubernetes", "bazel-bin", "build", "release-tars")
```
amwat, line 67: The build logic in general needs some refactoring, but let's fix it the other way around.
since we already have a separate bazel builder let's modify this one to have 
`target = "quick-release"`

pkg/build/release_push_build.go
```diff
@@ -18,36 +19,18 @@ var _ Stager = &ReleasePushBuild{}
 func (rpb *ReleasePushBuild) Stage(version string) error {
```
amwat, line 19: We want the version to be used somewhere.
Ideally this is exactly what the version should be named in the final bucket.

I only see a VersionSuffix capability. https://godoc.org/k8s.io/release/pkg/release#PushBuildOptions
Is there a way to override the entire name?

pkg/build/release_push_build.go
```diff
@@ -18,36 +19,18 @@ var _ Stager = &ReleasePushBuild{}
...
-	cmd.SetDir(cmdDir)
-	return cmd.Run()
+	return errors.Wrap(
+		release.NewPushBuild(&release.PushBuildOptions{
+			Bucket:         mat[1],
+			GCSSuffix:      mat[3],
+			AllowDup:       true,
+			CI:             mat[2] == "ci",
+			NoUpdateLatest: true,
+		}).Push(),
```
amwat, line 33: Let's also plumb through ImageLocation -> DockerRegistry similar to what the bazel stager does.

pkg/build/make.go
saschagrunert: Fixed as suggeset. :+1:

pkg/build/release_push_build.go
saschagrunert: Done, I set it to `""` for now otherwise it tries to push the images to a registry which I do not have permissions to push to.

saschagrunert:
Alrigtht, gave it another try with the changed build type:
```
> kubetest2 gke --repo-root . --build --strategy=make --stage=gs://kubernetes-release-gcb/ci
```
The build results can be found here: https://console.cloud.google.com/storage/browser/kubernetes-release-gcb/ci?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false

Another small fix for noupdatelatest is in flight there: https://github.com/kubernetes/release/pull/1638

pkg/build/release_push_build.go
amwat: The main use case for that is running tests against uncommited local changes. 
For now we can probably make suffix work, I assume the auto-determined version comes from `hack/print-workspace-status.sh`? we do something similar (in `sourceVersion()` ) and append a kubetest2 uuid suffix.
We can do something like

```
VersionSuffix: strings.TrimPrefix(version, sourceVersion(repoRoot))
```

saschagrunert: Ah, no I added an option to overwrite the version, I think this makes sense.

Ref: https://github.com/kubernetes/release/pull/1644

saschagrunert:
PTAL @amwat

pkg/build/release_push_build.go
```diff
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"regexp"
 
-	"sigs.k8s.io/kubetest2/pkg/exec"
+	"github.com/pkg/errors"
+	"k8s.io/release/pkg/build"
```
amwat, line 24: probably worth aliasing this since this package is also called build

amwat (commented):
Let's give this a shot!
nothing is currently using the make strategy. But I'll follow up this with changes to the gce build to start using this.

/lgtm
/approve
/hold

justaugustus (changes requested):
@saschagrunert -- Once https://github.com/kubernetes/release/pull/1731 merges, can you pull it in here?

The logic for handling GCS suffixes gets fixed in that one.
/hold

pkg/build/release_push_build.go
saschagrunert: Done :+1:

saschagrunert (commented):
> @saschagrunert -- Once [kubernetes/release#1731](https://github.com/kubernetes/release/pull/1731) merges, can you pull it in here?
> 
> The logic for handling GCS suffixes gets fixed in that one.
> /hold

Done :+1:

saschagrunert:
Had to rebase once again to include the latest krel fixes
/unhold

PR #62, opened 2020-09-26
Add a unique identifier to kubetest2 runs

amwat:
Provides a way for deployers and tester to store relevant artifacts for potential re-use.
Also, simplifies artifacts storage for different runs in local use cases.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -57,8 +57,8 @@ type Options interface {
 	// returns the path to the directory where artifacts should be written
 	// (including metadata files like junit_runner.xml)
 	ArtifactsDir() string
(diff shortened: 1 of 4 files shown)


pkg/types/types.go
```diff
@@ -57,6 +57,8 @@ type Options interface {
 	// returns the path to the directory where artifacts should be written
 	// (including metadata files like junit_runner.xml)
 	ArtifactsDir() string
+	// UUID returns a unique identifier for a kubetest2 run.
+	UUID() string
```
spiffxp, line 61: Would `RunID` be a more descriptive identifier?

amwat: done

spiffxp (approved):
/approve
/lgtm

PR #63, opened 2020-09-28
Add --release-channel flag for kubetest2 gke deployer

chizhg:
Since `--release-channel` is usually paired with `--cluster-version` flag, and there is a constraint for using them together - `--cluster-version` must be a valid version if used together with `--release-channel`, add `--release-channel` flag and perform necessary validations in constructing the `gcloud` command.

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/up.go
+++ b/kubetest2-gke/deployer/up.go
@@ -81,10 +81,18 @@ func (d *deployer) Up() error {
 					"--image-type="+image,
 					"--num-nodes="+strconv.Itoa(d.nodes),
 					"--network="+transformNetworkName(d.projects, d.network),
-					"--cluster-version="+d.Version,
 				)
 				if d.workloadIdentityEnabled {
 					args = append(args, fmt.Sprintf("--workload-pool=%s.svc.id.goog", project))
+				}
+				if d.ReleaseChannel != "" {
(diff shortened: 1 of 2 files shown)


kubetest2-gke/deployer/up.go
```diff
@@ -81,11 +81,19 @@ func (d *deployer) Up() error {
...
 				if d.workloadIdentityEnabled {
 					args = append(args, fmt.Sprintf("--workload-pool=%s.svc.id.goog", project))
 				}
+				if d.ReleaseChannel != "" {
+					args = append(args, "--release-channel="+d.ReleaseChannel)
+					// --cluster-version must be a specific valid version if --release-channel is not empty.
+					if d.Version != "" && d.Version != "latest" {
+						args = append(args, "--cluster-version="+d.Version)
+					}
+				} else if d.Version != "" {
```
amwat, line 94: the deployer itself doesn't allow empty string as the version because we want to be explicit about getting the latest version, and it's validated. so we can drop the empty string checks.

amwat (commented):
/approve

one comment:

kubetest2-gke/deployer/up.go
chizhg: Done. Thanks!

amwat (commented):
/lgtm

PR #64, opened 2020-09-29
[GKE] Add functionality to auto-generate cluster names

amwat:
This is especially useful in multi-cluster use cases where specifying cluster names explicitly is tedious and unnecessary.
Multi-project use cases will still need to set the names explicitly.

### DIFF
--- a/kubetest2-gke/deployer/up.go
+++ b/kubetest2-gke/deployer/up.go
@@ -242,7 +242,7 @@ func generateClusterNames(numClusters int, uid string) []string {
 		// Naming convention: https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#POSITIONAL-ARGUMENTS
 		// must start with an alphabet, max length 40
 
-		// 4 characters for kt2- prefix
+		// 4 characters for kt2- prefix (short for kubetest2)
 		const fixedClusterNamePrefix = "kt2-"
 		// 3 characters -99 suffix
 		clusterNameSuffix := strconv.Itoa(i)
--- a/kubetest2-gke/deployer/up.go
+++ b/kubetest2-gke/deployer/up.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


amwat:
/cc @chizhg 

FYI @mm4tt

kubetest2-gke/deployer/up.go
```diff
@@ -203,3 +204,76 @@ func (d *deployer) Kubeconfig() (string, error) {
...
+}
+
+func generateClusterNames(numClusters int, uid string) []string {
+	clusters := make([]string, numClusters)
+	for i := 1; i <= numClusters; i++ {
+		// Naming convention: https://cloud.google.com/sdk/gcloud/reference/container/clusters/create#POSITIONAL-ARGUMENTS
+		// must start with an alphabet, max length 40
+
+		// 4 characters for kt2- prefix
+		const fixedClusterNamePrefix = "kt2-"
```
chizhg, line 246: What does `kt2` stand for? Is it worth a comment?

kubetest2-gke/deployer/up.go
```diff
@@ -203,3 +204,76 @@ func (d *deployer) Kubeconfig() (string, error) {
...
+		if uid != "" {
+			const maxIDLength = 33
+			if len(uid) > maxIDLength {
+				id = uid[:maxIDLength]
+			} else {
+				id = uid
+			}
+			id += "-"
+		}
+		clusters[i-1] = fixedClusterNamePrefix + id + clusterNameSuffix
```
chizhg, line 260: Nit: this will be a long name, precisely 40 characters if the uid exists. For the sake of readability, would it be better to make it shorter?

kubetest2-gke/deployer/up.go
amwat: `kubetest2` ? added a clarifying comment

kubetest2-gke/deployer/up.go
amwat: I think we want to preserve as much of the unique identifier as possible for identifying different runs.
Since no one has to actually interact with the cluster names manually, the length of the cluster name shouldn't really be a problem? (40 is the max that GKE allows).

chizhg: SG, thanks!

PR #66, opened 2020-10-10
Allow specifying the latest version for a release channel

chizhg:
When we run `gcloud` command with `--release-channel=xxx --cluster-version=latest`, it will give an error like `ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=Master version must be one of "REGULAR" channel supported versions [1.16.13-gke.401, 1.17.9-gke.1504].`.

This PR allows users to specify `latest` as the cluster version when `--release-channel` is not empty by resolving the actual latest version from the server config.

Also allow specifying an empty string as the cluster version, in which case the default cluster version will be used.

### DIFF
--- a/kubetest2-gke/deployer/commandutils.go
+++ b/kubetest2-gke/deployer/commandutils.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/yaml.v2"
 	"k8s.io/klog"
 
 	"sigs.k8s.io/kubetest2/pkg/exec"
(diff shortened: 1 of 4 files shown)


kubetest2-gke/deployer/commandutils.go
```diff
@@ -116,3 +117,36 @@ func getClusterCredentials(project, loc, cluster string) error {
...
+	if err = yaml.Unmarshal(out, &cs); err != nil {
+		return "", fmt.Errorf("failed to unmarshal the server config: %w", err)
+	}
+
+	for _, channel := range cs.Channels {
+		if strings.EqualFold(channel.Name, channelName) {
+			if len(channel.ValidVersions) == 0 {
+				return "", fmt.Errorf("no valid versions for channel %q", channelName)
+			}
+			return channel.ValidVersions[0], nil
```
amwat, line 147: are the returned values already sorted in descending order?

kubetest2-gke/deployer/up.go
```diff
@@ -264,7 +271,7 @@ func generateClusterNames(numClusters int, uid string) []string {
 
 func validateVersion(version string) error {
 	switch version {
-	case "latest":
+	case "latest", "":
```
amwat, line 274: wanted to avoid this, but looks like the default version in the release channel is not the same as latest which is a different behavior from not using a release channel.

Let's also change the default flag value to empty string then.

amwat (commented):
/approve

kubetest2-gke/deployer/up.go
chizhg: Do you want me to change the default flag value to empty string for release channel only, or both? 

I did some quick experiment, the cluster version for `gcloud container clusters create test1 --cluster-version=` and `gcloud container clusters create test2 --cluster-version=latest` is actually different, the former is `1.16.13-gke.401` and the latter is `1.17.12-gke.500`.

kubetest2-gke/deployer/commandutils.go
chizhg: Yes, they are already sorted in descending order.

kubetest2-gke/deployer/up.go
amwat: Let's default to empty string for both cases.

chizhg: Done.

amwat (commented):
/lgtm

PR #68, opened 2020-10-21
Add helper for using exec.CommandContext

benmoss:
Really excited about this project! I am spiking on a CAPI provider (https://github.com/benmoss/kubetest2/tree/capi-deployer, though I think it might end up living in the core CAPI repo) and one of the things I needed was a version of the `ExecJUNit` process runner that takes a context.Context argument for doing command timeouts.

Thought it should live in this repo, since others probably can take advantage of it too.

### DIFF
--- a/pkg/process/junitexec.go
+++ b/pkg/process/junitexec.go
@@ -18,6 +18,7 @@ package process
 
 import (
 	"bytes"
+	"context"
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


amwat (commented):
/lgtm
/approve

Thanks!
Glad to hear that CAPI is adopting kubetest2!
FWIW making it possible to have out-of-tree deployers is one of the primary goals of this project so it's totally a good way to go.

PR #70, opened 2020-11-12
Normalize gke version

amwat:
Ensure that the generated version during build is a valid gke version.

/cc @MushuEE @liggitt

### DIFF
--- a/kubetest2-gke/deployer/build.go
+++ b/kubetest2-gke/deployer/build.go
@@ -18,6 +18,7 @@ package deployer
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 )
 
@@ -30,7 +31,11 @@ func (d *deployer) Build() error {
 		return err
 	}
 	version = strings.TrimPrefix(version, "v")
-	version += ".0+" + d.commonOptions.RunID()
+	if version, err = normalizeVersion(version); err != nil {
+		return err
+	}
+	// append the kubetest2 run id
+	version += "+" + d.commonOptions.RunID()
 	if d.BuildOptions.StageLocation != "" {
 		if err := d.BuildOptions.Stage(version); err != nil {
 			return fmt.Errorf("error staging build: %v", err)
@@ -46,4 +51,32 @@ func (d *deployer) verifyBuildFlags() error {
 	}
 	d.BuildOptions.RepoRoot = d.RepoRoot
 	return d.BuildOptions.Validate()
+}
+
+// ensure that the version is a valid gke version
+func normalizeVersion(version string) (string, error) {
+	// (1.18.10)(-gke)(.601.123)(+existingSuffix)
+	// capture repeated occurrences of group 3 and don't capture just last
+	re, err := regexp.Compile(`(\d\.\d+\.\d*)(-\w+)?((?:\.\d+)*)(\+.*)*`)
+	if err != nil {
+		return "", err
+	}
+	matches := re.FindStringSubmatch(version)
+	if matches == nil || len(matches) != 5 {
+		return "", fmt.Errorf("%q is not a valid gke version", version)
+	}
+	fmt.Println(matches)
+	// get the core version as is
+	finalVersion := matches[1]
+	// force append -gke, ignore alpha,beta
+	finalVersion += "-gke"
+	// add the optional patch number or .0
+	if matches[3] == "" {
+		finalVersion += ".0"
+	} else {
+		finalVersion += matches[3]
+	}
+	// append the optional suffix as is
+	finalVersion += matches[4]
+	return finalVersion, nil
 }
--- a/kubetest2-gke/deployer/build.go
+++ b/kubetest2-gke/deployer/build.go
@@ -20,6 +20,14 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+
+	"k8s.io/klog"
+)
+
+var (
+	// (1.18.10)(-gke)(.601.123)(+existingSuffix)
+	// capture repeated occurrences of group 3 and don't capture just last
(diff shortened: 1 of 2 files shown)


kubetest2-gke/deployer/build.go
```diff
@@ -47,3 +52,31 @@ func (d *deployer) verifyBuildFlags() error {
...
+	// capture repeated occurrences of group 3 and don't capture just last
+	re, err := regexp.Compile(`(\d\.\d+\.\d*)(-\w+)?((?:\.\d+)*)(\+.*)*`)
+	if err != nil {
+		return "", err
+	}
+	matches := re.FindStringSubmatch(version)
+	if matches == nil || len(matches) != 5 {
+		return "", fmt.Errorf("%q is not a valid gke version", version)
+	}
+	fmt.Println(matches)
```
liggitt, line 68: drop debug logging?

kubetest2-gke/deployer/build_test.go
```diff
@@ -0,0 +1,100 @@
...
+		{
+			name:            "alpha version",
+			version:         "1.16.13-alpha.123",
+			expectedVersion: "1.16.13-gke.123",
+			expectError:     false,
+		},
+		{
+			name:            "beta version, no minor, pre-existing suffix",
+			version:         "1.20.0-beta+qwe123",
+			expectedVersion: "1.20.0-gke.0+qwe123",
```
liggitt, line 79: would it make sense to just insert gke.0 and preserve all the other bits, e.g. `1.20.0-gke.0+beta+qwe123`?

kubetest2-gke/deployer/build.go
```diff
@@ -47,3 +52,31 @@ func (d *deployer) verifyBuildFlags() error {
...
+func normalizeVersion(version string) (string, error) {
+	// (1.18.10)(-gke)(.601.123)(+existingSuffix)
+	// capture repeated occurrences of group 3 and don't capture just last
+	re, err := regexp.Compile(`(\d\.\d+\.\d*)(-\w+)?((?:\.\d+)*)(\+.*)*`)
+	if err != nil {
+		return "", err
+	}
+	matches := re.FindStringSubmatch(version)
+	if matches == nil || len(matches) != 5 {
+		return "", fmt.Errorf("%q is not a valid gke version", version)
```
liggitt, line 66: include matches in error for debugging?
```suggestion
		return "", fmt.Errorf("%q is not a valid gke version: %#v", version, matches)
```

kubetest2-gke/deployer/build.go
```diff
@@ -47,3 +52,31 @@ func (d *deployer) verifyBuildFlags() error {
...
+	// force append -gke, ignore alpha,beta
+	finalVersion += "-gke"
+	// add the optional patch number or .0
+	if matches[3] == "" {
+		finalVersion += ".0"
+	} else {
+		finalVersion += matches[3]
+	}
+	// append the optional suffix as is
+	finalVersion += matches[4]
```
liggitt, line 80: if finalVersion != version, maybe print out that we transformed version to finalVersion?

kubetest2-gke/deployer/build.go
```diff
@@ -47,3 +52,31 @@ func (d *deployer) verifyBuildFlags() error {
 	d.BuildOptions.RepoRoot = d.RepoRoot
 	return d.BuildOptions.Validate()
 }
+
+// ensure that the version is a valid gke version
+func normalizeVersion(version string) (string, error) {
+	// (1.18.10)(-gke)(.601.123)(+existingSuffix)
+	// capture repeated occurrences of group 3 and don't capture just last
+	re, err := regexp.Compile(`(\d\.\d+\.\d*)(-\w+)?((?:\.\d+)*)(\+.*)*`)
```
MushuEE, line 60: Worth forcing as a constant with [mustCompile](https://golang.org/pkg/regexp/#MustCompile)? I forget where but someone had me use that on one of my PRs and will fail fast if a change breaks the regex.

kubetest2-gke/deployer/build.go
```diff
@@ -47,3 +52,31 @@ func (d *deployer) verifyBuildFlags() error {
...
+	if matches == nil || len(matches) != 5 {
+		return "", fmt.Errorf("%q is not a valid gke version", version)
+	}
+	fmt.Println(matches)
+	// get the core version as is
+	finalVersion := matches[1]
+	// force append -gke, ignore alpha,beta
+	finalVersion += "-gke"
+	// add the optional patch number or .0
+	if matches[3] == "" {
```
MushuEE, line 74: I started using [named regex](https://stackoverflow.com/questions/20750843/using-named-matches-from-go-regex) just because it was so much nicer for me to track over indexes.

kubetest2-gke/deployer/build.go
amwat: done

kubetest2-gke/deployer/build.go
amwat: TIL, thanks. done.

kubetest2-gke/deployer/build.go
amwat: done

kubetest2-gke/deployer/build.go
amwat: done

kubetest2-gke/deployer/build_test.go
amwat: makes sense, moved it as a suffix and made it `.99.0` instead. single patch didn't seem to work for me.

kubetest2-gke/deployer/build.go
amwat: done, thanks for catching.

PR #71, opened 2020-11-13
Update GKE version extraction/normalization

liggitt:
Follow up to https://github.com/kubernetes-sigs/kubetest2/pull/70 with additional tests/matchers

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/build_test.go
+++ b/kubetest2-gke/deployer/build_test.go
@@ -38,9 +38,51 @@ func TestNormalizeVersion(t *testing.T) {
 			expectError:     false,
 		},
 		{
-			name:            "full version",
+			name:            "oss rc tag",
+			version:         "1.19.4-rc.0",
+			expectedVersion: "1.19.4-gke.99.0+rc.0",
+			expectError:     false,
+		},
+		{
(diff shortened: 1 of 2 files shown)


kubetest2-gke/deployer/build_test.go
```diff
@@ -70,7 +112,7 @@ func TestNormalizeVersion(t *testing.T) {
 		{
 			name:            "alpha version with patch",
 			version:         "1.16.13-alpha.123",
-			expectedVersion: "1.16.13-gke.123+alpha",
+			expectedVersion: "1.16.13-gke.99.0+alpha.123",
```
amwat, line 115: This behavior is changing, are we sure we don't want `.123.0` instead?

amwat (commented):
one question, lgtm otherwise

/lgtm
/approve
/hold

kubetest2-gke/deployer/build_test.go
liggitt: yeah, alpha.123 has no real bearing on gke build version... preserving the existing suffix intact makes more sense to me

amwat: sgtm

/hold cancel

PR #73, opened 2020-11-17
Fix --private-cluster-master-ip-range arg for multicluster

chizhg:
Fix `--private-cluster-master-ip-range` arg for multicluster: 

when we create multiple private clusters, we need to specify different `master-ip-range` for each cluster, otherwise there will be conflict. This PR allows the `--private-cluster-master-ip-range` arg to be accepting an array, and the length of it must be the same as the number of clusters.

### DIFF
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -79,6 +79,12 @@ type ig struct {
 	uniq string
 }
 
+type cluster struct {
+	// index is the index of the cluster in the list provided via the --cluster-name flag
+	index int
+	name  string
+}
+
 type deployer struct {
 	// generic parts
 	commonOptions types.Options
@@ -98,7 +104,7 @@ type deployer struct {
 	region   string
 	clusters []string
 	// only used for multi-project multi-cluster profile to save the project-clusters mapping
-	projectClustersLayout map[string][]string
+	projectClustersLayout map[string][]cluster
 	nodes                 int
 	machineType           string
 	network               string
@@ -124,8 +130,8 @@ type deployer struct {
 
 	// Private cluster access level, must be one of "no", "limited" and "unrestricted".
 	// See the details in https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters
-	privateClusterAccessLevel   string
-	privateClusterMasterIPRange string
+	privateClusterAccessLevel    string
+	privateClusterMasterIPRanges []string
 
 	boskosLocation              string
 	boskosResourceType          string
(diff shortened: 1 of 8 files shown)


kubetest2-gke/deployer/up.go
```diff
@@ -65,12 +65,14 @@ func (d *deployer) Up() error {
...
 	eg, ctx := errgroup.WithContext(ctx)
 	loc := location(d.region, d.zone)
+	clusterIndex := 0
 	for i := range d.projects {
 		project := d.projects[i]
 		subNetworkArgs := subNetworkArgs(d.projects, d.region, d.network, i)
 		for j := range d.projectClustersLayout[project] {
 			cluster := d.projectClustersLayout[project][j]
-			privateClusterArgs := privateClusterArgs(d.network, cluster, d.privateClusterAccessLevel, d.privateClusterMasterIPRange)
+			privateClusterArgs := privateClusterArgs(d.network, cluster, d.privateClusterAccessLevel, d.privateClusterMasterIPRanges[clusterIndex])
```
amwat, line 74: do we want to retain the 1:1 mapping of the clusters input and private master range input since they are tightly coupled?
currently it would assign the private ip range based the iteration order of the map.

kubetest2-gke/deployer/up.go
```diff
@@ -217,9 +219,7 @@ func (d *deployer) verifyUpFlags() error {
 	if len(d.projects) == 0 && d.boskosProjectsRequested <= 0 {
 		return fmt.Errorf("either --project or --projects-requested with a value larger than 0 must be set for GKE deployment")
 	}
-	if err := d.verifyNetworkFlags(); err != nil {
-		return err
-	}
+
 	if len(d.clusters) == 0 {
 		if len(d.projects) > 1 || d.boskosProjectsRequested > 1 {
 			return fmt.Errorf("explicit --cluster-name must be set for multi-project profile")
```
amwat, line 225: Let's move these checks above as well.

amwat (commented):
/approve

kubetest2-gke/deployer/up.go
chizhg: I have updated the logic to retain the 1:1 mapping. Thanks!

kubetest2-gke/deployer/up.go
chizhg: Hmm not sure if I got your point, could you elaborate?

amwat: nvm, I think I saw the diff wrong. you **are** moving the length check above verifyNetworkFlags

kubetest2-gke/deployer/deployer.go
```diff
@@ -79,6 +79,12 @@ type ig struct {
 	uniq string
 }
 
+type cluster struct {
+	// index is the index of the cluster in the list provided via the --cluster-name flag
+	index int
+	name  string
+}
```
amwat, line 86: this is a good start, we should continue to move other attributes inside this struct, to construct an object with all relevant creation info.

amwat (commented):
/lgtm
/approve

PR #75, opened 2020-11-23
Fix an 'index out of range' error when creating non-private clusters

chizhg:
When creating multiple non-private clusters, the length of `privateClusterMasterIPRanges` will be default to 1, but the cluster index can be larger than 0, thus causing an 'index out of range' error.

This PR fixes the bug.

/cc @amwat

### DIFF
--- a/kubetest2-gke/deployer/network.go
+++ b/kubetest2-gke/deployer/network.go
@@ -362,18 +362,18 @@ func removeHostServiceAgentUserRole(projects []string) error {
 
 // This function returns the args required for creating a private cluster.
 // Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#top_of_page
-func privateClusterArgs(network, cluster, accessLevel, masterIPRange string) []string {
+func privateClusterArgs(network, accessLevel string, masterIPRanges []string, clusterInfo cluster) []string {
 	if accessLevel == "" {
 		return []string{}
 	}
 
-	subnetName := network + "-" + cluster
+	subnetName := network + "-" + clusterInfo.name
 	common := []string{
 		"--create-subnetwork=name=" + subnetName,
 		"--enable-ip-alias",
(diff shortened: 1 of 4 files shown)


kubetest2-gke/deployer/network.go
```diff
@@ -362,18 +362,18 @@ func removeHostServiceAgentUserRole(projects []string) error {
 
 // This function returns the args required for creating a private cluster.
 // Reference: https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters#top_of_page
-func privateClusterArgs(network, cluster, accessLevel, masterIPRange string) []string {
+func privateClusterArgs(network, accessLevel string, masterIPRanges []string, clusterInfo cluster) []string {
```
amwat, line 365: let's add some unit tests as sanity checks to prevent future issues.

chizhg: Added, thanks!

kubetest2-gke/deployer/network_test.go
```diff
@@ -0,0 +1,103 @@
...
+		accessLevel    string
+		masterIPRanges []string
+		clusterInfo    cluster
+		expected       []string
+	}{
+		{
+			desc:           "no private cluster args are needed for non-private clusters",
+			network:        "whatever-network",
+			accessLevel:    "",
+			masterIPRanges: []string{"whatever-master-IP-range"},
```
amwat, line 38: nit: for non private clusters, we need a (new/modify this) case where masterIPRanges could be empty.

amwat (commented):
/approve

kubetest2-gke/deployer/network_test.go
chizhg: Added the test case, thanks

amwat (commented):
/lgtm

PR #78, opened 2020-12-21
[GCE] use kubectl from PATH instead of get-kube-binaries

amwat:
Depending on get-kube-binaries is unreliable 
https://github.com/kubernetes/kubernetes/blob/b10d82b93bad7a4e39b9d3f5c5e81defa3af68f0/cluster/kubectl.sh#L25-L26
and leading to failures in 
https://testgrid.k8s.io/conformance-gce#Conformance%20-%20GCE%20-%20master%20-%20kubetest2

instead rely on kubectl existing in $PATH, that is the recommended approach for local testing as mentioned above
kubectl already exists in all the CI images [kubekins-e2e](https://github.com/kubernetes/test-infra/blob/6d5079cd601273d85d9c57a6eccdfedc853d8376/images/kubekins-e2e/Dockerfile#L59-L64) and [krte](https://github.com/kubernetes/test-infra/blob/6d5079cd601273d85d9c57a6eccdfedc853d8376/images/krte/Dockerfile#L87-L96)

/cc @MushuEE 

fyi @krzykwas


MushuEE (commented):
/lgtm

BenTheElder:
> instead rely on kubectl existing in $PATH, that is the recommended approach for local testing as mentioned above
kubectl already exists in all the CI images kubekins-e2e and krte

When testing a particular kubernetes version we should be using the kubectl from that version, the one in the CI image is going to lag behind (it comes from GKE, which has the magic version matcher, but only so many versions).

BenTheElder:
IMHO the answer to that though is for the deployer to inject the new kubectl to the front of PATH something like put it in an otherwise empty directory and then prepend that to PATH.

PR #80, opened 2021-01-13
Unify common build functionality between gce and gke

amwat:
Also extend make build support for both and start staging with the new krel API.

/cc @MushuEE @BenTheElder @spiffxp


kubetest2-gce/deployer/options/build.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
MushuEE, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

pkg/build/make.go
```diff
@@ -18,101 +18,30 @@ package build
...
-	if err != nil {
-		return "", err
-	}
-	version, err := sourceVersion(src)
+	version, err := sourceVersion(m.RepoRoot)
 	if err != nil {
 		return "", fmt.Errorf("failed to get version: %v", err)
 	}
-	cmd := exec.Command("make", "-C", src, target)
+	cmd := exec.Command("make", "-C", m.RepoRoot, target)
```
MushuEE, line 41: Just checking, the Make file exists at Root?

pkg/build/release_push_build.go
```diff
@@ -18,28 +18,47 @@ package build
...
 func (rpb *ReleasePushBuild) Stage(version string) error {
+	if !strings.HasPrefix(version, "v") {
+		version = "v" + version
+	}
 	re := regexp.MustCompile(`^gs://([\w-]+)/(devel|ci)(/.*)?`)
-	mat := re.FindStringSubmatch(rpb.Location)
+	mat := re.FindStringSubmatch(rpb.StageLocation)
 	if mat == nil || len(mat) < 4 {
-		return fmt.Errorf("invalid stage location: %v. Use gs://bucket/ci/optional-suffix", rpb.Location)
+		return fmt.Errorf("invalid stage location: %v. Use gs://bucket/ci/optional-suffix", rpb.StageLocation)
```
MushuEE, line 48:
```suggestion
		return fmt.Errorf("invalid stage location: %v. Use form gs://<bucket>/<ci>/<optional-suffix>", rpb.StageLocation)
```

pkg/build/release_push_build.go
```diff
@@ -18,28 +18,47 @@ package build
...
+	}
+
+	// force indicate staging of extra gce configuration files
```
MushuEE, line 51:
```suggestion
	// Force indicate staging of extra gce configuration files
```

kubetest2-gke/deployer/options/build.go
```diff
@@ -17,41 +17,24 @@ limitations under the License.
...
 type BuildOptions struct {
-	Strategy      string `flag:"~strategy" desc:"Determines the build strategy to use either make or bazel."`
-	StageLocation string `flag:"~stage" desc:"Upload binaries to gs://bucket/ci/job-suffix if set"`
-	RepoRoot      string `flag:"-"`
-	build.Builder
-	build.Stager
+	CommonBuildOptions *build.Options
 }
 
+var _ build.Builder = &BuildOptions{}
```
MushuEE, line 27: Curious what the propose of this pattern is? Just type enforcement?

pkg/build/make.go
amwat: yup, https://github.com/kubernetes/kubernetes/blob/master/Makefile

kubetest2-gke/deployer/options/build.go
amwat: yes, this essentially is a compile time type assertion

pkg/build/make.go
BenTheElder: it's actually a symlink, but yes 🙃 
the real file is in ./build/root/Makefile

kubetest2-gce/deployer/build.go
```diff
@@ -34,14 +34,17 @@ func (d *deployer) Build() error {
 
 	if d.LegacyMode {
 		// this supports the kubernetes/kubernetes build
-		klog.V(2).Info("starting the legacy build")
-
-		cmd := exec.Command("make", "bazel-release")
-		exec.InheritOutput(cmd)
-		cmd.SetDir(d.RepoRoot)
-		err := cmd.Run()
+		klog.V(2).Info("starting the legacy k/k build")
```
spiffxp, line 37:
```suggestion
		klog.V(2).Info("starting the legacy kubernetes/kubernetes build")
```

kubetest2-gce/deployer/build.go
```diff
@@ -34,14 +34,17 @@ func (d *deployer) Build() error {
...
+		version, err := d.BuildOptions.Build()
 		if err != nil {
-			return fmt.Errorf("error during make step of build: %s", err)
+			return err
+		}
+		// append the kubetest2 run id
+		version += "+" + d.commonOptions.RunID()
+		if d.BuildOptions.CommonBuildOptions.StageLocation != "" {
```
spiffxp, line 44: nit
```suggestion
		version += "+" + d.commonOptions.RunID()
		// stage build if requested
		if d.BuildOptions.CommonBuildOptions.StageLocation != "" {
```

I'm a little curious why stage needs to be done here, seems like a side effect unrelated to building, but won't block

kubetest2-gke/deployer/build.go
```diff
@@ -55,7 +67,7 @@ func (d *deployer) Build() error {
 	}
 	// append the kubetest2 run id
 	version += "+" + d.commonOptions.RunID()
-	if d.BuildOptions.StageLocation != "" {
+	if d.BuildOptions.CommonBuildOptions.StageLocation != "" {
```
spiffxp, line 70:
```suggestion
        // stage build if requested
	if d.BuildOptions.CommonBuildOptions.StageLocation != "" {
```

pkg/build/options.go
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+package build
+
+import (
+	"fmt"
+)
+
+type Options struct {
+	Strategy           string `flag:"~strategy" desc:"Determines the build strategy to use either make or bazel."`
```
spiffxp, line 24: nit: seems like this could be its own type, but won't block

spiffxp (changes requested):
tiny suggestions otherwise LGTM

kubetest2-gce/deployer/build.go
amwat: yeah stage is optional here for gce, it's not needed if running build and up together. but needed if we want to build and then up a cluster later.

pkg/build/options.go
amwat: They are separate types: 
https://github.com/kubernetes-sigs/kubetest2/blob/03dbeda27f3fc7a7629d6a01e12e8174eceee6a5/pkg/build/make.go#L25-L27

https://github.com/kubernetes-sigs/kubetest2/blob/03dbeda27f3fc7a7629d6a01e12e8174eceee6a5/pkg/build/bazel.go#L27-L31

the string here is to expose the available strategies as a flag.

kubetest2-gke/deployer/build.go
BenTheElder: done, I think?

kubetest2-gke/deployer/options/build.go
BenTheElder: adding: this is a common enough pattern in go, I'm not sure if it's strictly compile time or not though. 

technically I bet at minimum it also ensures that the package defining the interface is valid to import even if it otherwise wouldn't be, and importing a package can have side effects (namely packages have `init()` / globals in a binary IFF they are imported somewhere).

kubetest2-gke/deployer/options/build.go
```diff
@@ -17,41 +17,26 @@ limitations under the License.
...
+var _ build.Stager = &BuildOptions{}
+
 func (bo *BuildOptions) Validate() error {
-	return bo.implementationFromStrategy()
+	// force extra GCP files to be staged
```
BenTheElder, line 31: nit: if we're always going to force this, IMHO this should be in `Build()` or really anywhere other than here, `Validate` should normally be only reading data, not writing; this fails the principle of least surprise.

pkg/build/options.go
BenTheElder: most flag libraries allow you to use the real type instead of a string /shrug

BenTheElder: the upside to doing so is that you can then pass around this struct as the real parsed values

kubetest2-gke/deployer/options/build.go
amwat: Done. also added a check to ensure stage is used with build and up since that's required for gke.

pkg/build/options.go
amwat: we are using the real type which is `BuildOptions` and has the actual options like RepoRoot , but at some point you need to specify whether it's bazel or make :)

exposing the individual implementation directly would mean weirdly exposing individual flags like --make-repo-root and --bazel-repo-root which looks worse /shrug

amwat:
ping

kubetest2-gke/deployer/build.go
```diff
@@ -39,9 +40,20 @@ var (
 	// $1 == prefix
 	// $2 == suffix
 	buildPrefix = regexp.MustCompile(`^(\d\.\d+\.\d+)([+-].*)?$`)
+
+	defaultImageTag = "gke.gcr.io"
```
spiffxp, line 44: nit (can be done as followup), seems like `defaultImageRegistry` would be a better name

kubetest2-gke/deployer/build.go
```diff
@@ -39,9 +40,20 @@ var (
...
 )
 
 func (d *deployer) Build() error {
+	imageTag := defaultImageTag
+	if d.BuildOptions.CommonBuildOptions.ImageLocation != "" {
+		imageTag = d.BuildOptions.CommonBuildOptions.ImageLocation
+	}
+
+	klog.V(2).Infof("setting KUBE_DOCKER_REGISTRY to %s for tagging images", imageTag)
+	if err := os.Setenv("KUBE_DOCKER_REGISTRY", imageTag); err != nil {
```
spiffxp, line 54: ditto for `imageRegistry` instead

pkg/build/options.go
spiffxp: I guess I could have been clearer and suggested a string-as-enum approach (`type BuildStrategy string` etc.) but as-is is fine, like I said, won't block.

spiffxp (approved):
/approve
/lgtm

spiffxp:
/test gce-build-up-down
```
 - The zone 'projects/k8s-e2e-gce-alpha1-5/zones/us-central1-b' does not have enough resources available to fulfill the request.  Try a different zone, or try again later.
```

Neat.

kubetest2-gce/deployer/build.go
```diff
@@ -34,14 +34,18 @@ func (d *deployer) Build() error {
...
-		cmd.SetDir(d.RepoRoot)
-		err := cmd.Run()
+		klog.V(2).Info("starting the legacy kubernetes/kubernetes build")
+		version, err := d.BuildOptions.Build()
 		if err != nil {
-			return fmt.Errorf("error during make step of build: %s", err)
+			return err
+		}
+		// append the kubetest2 run id
+		version += "+" + d.commonOptions.RunID()
```
liggitt, line 43: if version already contains "+", this should append `"+" + d.commonOptions.RunID()`, otherwise we end up with invalid docker tags

amwat: this is the GCE deployer not GKE, we might want similar changes here too though. yet to test this out for upstream jobs.

liggitt: the double-plus version results in an invalid docker tag regardless of deployer

PR #82, opened 2021-01-15
Fix kind deployer build type

bobbypage:
Building a kind node image with bazel wasn't working since the
`--build-type` flag was reading the nodeImage argument incorrectly. This
fixes the flag parsing to ensure the result of `--build-type` will be
correctly stored in deployer's buildType field.


amwat (commented):
/lgtm
/approve

Thanks!

PR #83, opened 2021-01-15
Expose the ginkgo test package marker as a flag

rifelpet:
This allows `latest.txt` to be overridden to support testing release branches with `latest-1.20.txt` for example.


amwat (commented):
/lgtm
/approve

Thanks!

PR #84, opened 2021-01-19
Log the commands before running them for easier debugging

chizhg:
Sometimes we want to debug issues related to `kubetest2` but it's pretty hard to dig out the commands that were run during the process. This PR will log the commands before running them for easier debugging.

### DIFF
--- a/pkg/exec/local.go
+++ b/pkg/exec/local.go
@@ -20,6 +20,9 @@ import (
 	"context"
 	"io"
 	osexec "os/exec"
+	"strings"
+
+	"k8s.io/klog"
 )
 
 // LocalCmd wraps os/exec.Cmd, implementing the exec.Cmd interface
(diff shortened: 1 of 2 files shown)


pkg/exec/local.go
```diff
@@ -36,13 +39,15 @@ var _ Cmder = &LocalCmder{}
...
 func (c *LocalCmder) Command(name string, arg ...string) Cmd {
+	klog.V(2).Infof("⚙️ %s %s", name, strings.Join(arg, " "))
 	return &LocalCmd{
 		Cmd: osexec.Command(name, arg...),
 	}
 }
 
 // CommandContext returns a new exec.Cmd with the context, backed by Cmd
 func (c *LocalCmder) CommandContext(ctx context.Context, name string, arg ...string) Cmd {
+	klog.V(2).Infof("⚙️ %s %s", name, strings.Join(arg, " "))
```
amwat, line 50: This might cause issues rendering in some terminals, but we can add handling for `TERM=dumb` later if we run into any issues.

amwat (commented):
/lgtm
/approve

PR #85, opened 2021-01-25
Allow setting backup regions and zones when creating GKE clusters and a few other fixes

chizhg:
This PR includes a few changes:

1. Allow setting backup regions and zones when creating GKE clusters. And when specific errors happen, retry creating the clusters in these backup regions/zones. This is important since a lot of infra flakes are caused by these errors, and there is not a good way to prevent them except retrying, see more details in the internal issue 162609408

2. Change the log level for logging the commands from `2` to `1`, since they are more important than the level `2` logging

3. Fix a bug when running kubetest2-gke only with `--down` but not `--up` flag

4. A few other small coding style improvements and cleanups

### DIFF
--- a/kubetest2-gke/deployer/commandutils.go
+++ b/kubetest2-gke/deployer/commandutils.go
@@ -17,8 +17,11 @@ limitations under the License.
 package deployer
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"os"
+	osexec "os/exec"
 	"path/filepath"
 	"strings"
 
@@ -49,7 +52,7 @@ func (d *deployer) prepareGcpIfNeeded(projectID string) error {
 	}
 
 	if err := os.Setenv("CLOUDSDK_CORE_PRINT_UNHANDLED_TRACEBACKS", "1"); err != nil {
-		return fmt.Errorf("could not set CLOUDSDK_CORE_PRINT_UNHANDLED_TRACEBACKS=1: %v", err)
+		return fmt.Errorf("could not set CLOUDSDK_CORE_PRINT_UNHANDLED_TRACEBACKS=1: %w", err)
 	}
 	if err := os.Setenv("CLOUDSDK_API_ENDPOINT_OVERRIDES_CONTAINER", endpoint); err != nil {
 		return err
@@ -112,7 +115,7 @@ func getClusterCredentials(project, loc, cluster string) error {
 	if err := runWithOutput(exec.Command("gcloud",
 		containerArgs("clusters", "get-credentials", cluster, "--project="+project, loc)...),
 	); err != nil {
-		return fmt.Errorf("error executing get-credentials: %v", err)
+		return fmt.Errorf("error executing get-credentials: %w", err)
 	}
 
 	return nil
@@ -149,4 +152,37 @@ func resolveLatestVersionInChannel(loc, channelName string) (string, error) {
 	}
 
 	return "", fmt.Errorf("channel %q does not exist in the server config", channelName)
+}
+
+func containerArgs(args ...string) []string {
+	return append(append([]string{}, "container"), args...)
+}
+
+func runWithNoOutput(cmd exec.Cmd) error {
+	exec.NoOutput(cmd)
+	return cmd.Run()
+}
+
+func runWithOutput(cmd exec.Cmd) error {
+	exec.InheritOutput(cmd)
+	return cmd.Run()
+}
+
+func runWithOutputAndBuffer(cmd exec.Cmd) (string, error) {
+	var buf bytes.Buffer
+
+	exec.SetOutput(cmd, io.MultiWriter(os.Stdout, &buf), io.MultiWriter(os.Stderr, &buf))
+	if err := cmd.Run(); err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+// execError returns a string format of err including stderr if the
+// err is an ExitError, useful for errors from e.g. exec.Cmd.Output().
+func execError(err error) string {
+	if ee, ok := err.(*osexec.ExitError); ok {
+		return fmt.Sprintf("%v (output: %q)", err, string(ee.Stderr))
+	}
+	return err.Error()
 }
--- a/kubetest2-gke/deployer/up.go
+++ b/kubetest2-gke/deployer/up.go
@@ -18,6 +18,7 @@ package deployer
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -32,6 +33,17 @@ import (
 	"sigs.k8s.io/kubetest2/pkg/exec"
 	"sigs.k8s.io/kubetest2/pkg/metadata"
 )
+
+// If one of the error patterns below is matched, it would be recommended to
+// retry creating the cluster in a different region.
+// - stockout
+// - nodes fail to start
+// - component is unhealthy
+var retryableCreationErrors = []*regexp.Regexp{
+	regexp.MustCompile(`.*does not have enough resources available to fulfill.*`),
+	regexp.MustCompile(`.*only \d+ nodes out of \d+ have registered; this is likely due to Nodes failing to start correctly.*`),
+	regexp.MustCompile(`.*All cluster resources were brought up, but: component .+ from endpoint .+ is unhealthy.*`),
+}
 
 // Deployer implementation methods below
 func (d *deployer) Up() error {
@@ -53,6 +65,53 @@ func (d *deployer) Up() error {
 	if err := d.prepareGcpIfNeeded(d.projects[0]); err != nil {
 		return err
 	}
+
+	var err error
+	if d.region != "" {
+		for i, r := range append([]string{d.region}, d.backupRegions...) {
+			if i != 0 {
+				klog.V(2).Infof("Retrying creating the clusters in the backup region %q", r)
+			}
+			d.region = r
+
+			if err = d.createClusters(); err == nil || !isRetryableError(err) {
+				break
+			}
+		}
+	} else {
+		for i, z := range append([]string{d.zone}, d.backupZones...) {
+			if i != 0 {
+				klog.V(2).Infof("Retrying creating the clusters in the backup zone %q", z)
+			}
+			d.zone = z
+
+			if err = d.createClusters(); err == nil || !isRetryableError(err) {
+				break
+			}
+		}
+	}
+	if err != nil {
+		return err
+	}
+
+	if err = d.testSetup(); err != nil {
+		return fmt.Errorf("error running setup for the tests: %w", err)
+	}
+
+	return nil
+}
+
+// isRetryableError checks if the error happens during cluster creation can be potentially solved by retrying or not.
+func isRetryableError(err error) bool {
+	for _, regx := range retryableCreationErrors {
+		if regx.MatchString(err.Error()) {
+			return true
+		}
+	}
+	return false
+}
+
+func (d *deployer) createClusters() error {
 	if err := d.createNetwork(); err != nil {
 		return err
 	}
@@ -61,13 +120,14 @@ func (d *deployer) Up() error {
 	}
 
 	klog.V(2).Infof("Environment: %v", os.Environ())
+
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 	eg, ctx := errgroup.WithContext(ctx)
-	loc := location(d.region, d.zone)
+	loc := locationFlag(d.region, d.zone)
 	for i := range d.projects {
 		project := d.projects[i]
-		subNetworkArgs := subNetworkArgs(d.projects, d.region, d.network, i)
+		subNetworkArgs := subNetworkArgs(d.projects, d.region, d.zone, d.network, i)
 		for j := range d.projectClustersLayout[project] {
 			cluster := d.projectClustersLayout[project][j]
 			privateClusterArgs := privateClusterArgs(d.network, d.privateClusterAccessLevel, d.privateClusterMasterIPRanges, cluster)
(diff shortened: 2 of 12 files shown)


kubetest2-gke/deployer/commandutils.go
```diff
@@ -17,8 +17,10 @@ limitations under the License.
 package deployer
 
 import (
+	"bytes"
 	"fmt"
 	"os"
+	realexec "os/exec"
```
amwat, line 23: `osexec` would probably be a more appropriate alias here

pkg/exec/exec.go
```diff
@@ -116,6 +116,12 @@ func CombinedOutputLines(cmd Cmd) (lines []string, err error) {
 	return lines, err
 }
 
+// InheritOutputWithBuf sets cmd's output to write to the current process's stdout and stderr, and also writes the combined output to a buffer.
+func InheritOutputWithBuf(cmd Cmd, buf *bytes.Buffer) {
```
amwat, line 120: buffer seems oddly specific, we should make it io.Writer instead?

kubetest2-gke/deployer/up.go
```diff
@@ -105,22 +166,19 @@ func (d *deployer) Up() error {
 				args = append(args, subNetworkArgs...)
 				args = append(args, privateClusterArgs...)
 				args = append(args, cluster.name)
-				if err := runWithOutput(exec.CommandContext(ctx, "gcloud", args...)); err != nil {
+				var buf bytes.Buffer
+				if err := runWithOutputAndBuffer(exec.CommandContext(ctx, "gcloud", args...), &buf); err != nil {
 					// Cancel the context to kill other cluster creation processes if any error happens.
 					cancel()
-					return fmt.Errorf("error creating cluster: %v", err)
+					return fmt.Errorf("error creating cluster %q: %w", cluster.name, errors.New(buf.String()))
```
amwat, line 173: won't this buffer also have stdout ?

kubetest2-gke/deployer/up.go
```diff
@@ -53,6 +66,53 @@ func (d *deployer) Up() error {
...
+			}
+		}
+	} else {
+		for i, z := range append([]string{d.zone}, d.backupZones...) {
+			if i != 0 {
+				klog.V(2).Infof("Retrying creating the clusters in the backup zone %q", z)
+			}
+			d.zone = z
+
+			if err = d.createClusters(); err == nil || !isRetryableError(err) {
```
amwat, line 89: this messes "down". we would need to delete resources from failed attempts as well.

amwat (commented):
I feel like this is the wrong layer to solve stockout issues with fallbacks.
a single kubetest2 invocation should ideally be reproducible.

You can probably fix this at the job level right? by invoking multiple kubetest2 runs in different zones?

chizhg:
> I feel like this is the wrong layer to solve stockout issues with fallbacks.
> a single kubetest2 invocation should ideally be reproducible.
> 
> You can probably fix this at the job level right? by invoking multiple kubetest2 runs in different zones?

This is also my initial thought on this, and we already did it in Knative - https://github.com/knative/test-infra/blob/master/pkg/clustermanager/kubetest2/gke.go#L47. 

But we are also seeing this issues in Istio, which is the most common infra failures we've seen in the past few months, and I feel it's common to all the projects that require frequent cluster creation, so I think it makes more sense to make the update in upstream instead of duplicating the logic everywhere.

WDYT?

chizhg:
> I feel like this is the wrong layer to solve stockout issues with fallbacks.
> a single kubetest2 invocation should ideally be reproducible.
> 
> You can probably fix this at the job level right? by invoking multiple kubetest2 runs in different zones?

I'm not sure if I correctly understood what you meant by `invoking multiple kubetest2 runs in different zones` - but we cannot predict the `stockout` errors, since it only happens when there are not enough machines/resources in that data center to create the requested nodes. So I think `catch-failure-and-retry` is the only pattern to mitigate this issue.

kubetest2-gke/deployer/commandutils.go
chizhg: Done.

kubetest2-gke/deployer/up.go
```diff
@@ -33,6 +35,17 @@ import (
...
 
+// If one of the error patterns below is matched, it would be recommended to
+// retry creating the cluster in a different region.
+// - stockout
+// - nodes fail to start
+// - component is unhealthy
+var retryableCreationErrors = []*regexp.Regexp{
+	regexp.MustCompile(`.*does not have enough resources available to fulfill.*`),
+	regexp.MustCompile(`.*only \d+ nodes out of \d+ have registered; this is likely due to Nodes failing to start correctly.*`),
+	regexp.MustCompile(`.*All cluster resources were brought up, but: component .+ from endpoint .+ is unhealthy.*`),
```
amwat, line 46: component unhealthy doesn't seem like a retryable error. we see similar issues many times for legitimate cluster bring up issues.

kubetest2-gke/deployer/up.go
```diff
@@ -33,6 +35,17 @@ import (
 	"sigs.k8s.io/kubetest2/pkg/metadata"
 )
 
+// If one of the error patterns below is matched, it would be recommended to
+// retry creating the cluster in a different region.
+// - stockout
+// - nodes fail to start
+// - component is unhealthy
+var retryableCreationErrors = []*regexp.Regexp{
+	regexp.MustCompile(`.*does not have enough resources available to fulfill.*`),
```
amwat, line 44: we should differentiate stockout errors from regular quota errors.

same concerns as: https://github.com/kubernetes-sigs/kubetest2/pull/13#issuecomment-668292386

amwat (commented):
yeah, I meant the same retry logic that you have done for knative.
how do we plan on handling multiple clusters? if one fails do we recreate all of them in new zones or just the one that failed?
we also need to take care of cleanup for all creation attempts.

kubetest2-gke/deployer/up.go
chizhg: I don't have a good solution to this comment, but IMO we will only need `--backup-regions` and `--backup-zones` in CI since the flakiness is not really important if the users want to run the test flows on their local. And in CI we always have Boskos janitor to clean them up so it seems not a big issue.

Also the `down` flow is already broken since now it will throw an error when deleting clusters, if `up` failed to create the clusters due to quota issues, in which the networking and firewall rules will be leaked.

kubetest2-gke/deployer/up.go
chizhg: Do you mean it won't succeed even if we retry in different regions/zones? And does it mean we'll have to allow the infra flakiness caused by this error?

kubetest2-gke/deployer/up.go
chizhg: Ditto. The same reply as https://github.com/kubernetes-sigs/kubetest2/pull/85#discussion_r564119419

kubetest2-gke/deployer/up.go
chizhg: Yeah the output will be written to both stdout and the buffer.

The reasons I'm writing like this are:
1. We need to capture the output for matching the error pattern to decide if it needs to be retried
2. We still need to keep the stdout and stderr for informational purpose

Does it make sense to you?

pkg/exec/exec.go
chizhg: Replied in the other comment. We cannot use `io.Writer` here since we need to capture the output as well as printed in the stdout and stderr. Or do you mean `io.MultiWriter`?

chizhg:
To add more clarifications to the errors we want to retry here:

1. `.*does not have enough resources available to fulfill.*` corresponds to the `GCE_STOCKOUT` error. As mentioned in https://github.com/kubernetes-sigs/kubetest2/pull/85#issuecomment-767141860, there is no way to prevent this error, and the only way to mitigate will be retrying the cluster creation in a different region/zone.

2. `.*All cluster resources were brought up, but: component .+ from endpoint .+ is unhealthy.*` is probably a legitimate cluster bring up issue as you mentioned, but it is the reason for a large portion of infra flakiness, and based on our experience on Knative, retrying the cluster creation in a different region/zone will usually succeed.

3. `.*only \d+ nodes out of \d+ have registered; this is likely due to Nodes failing to start correctly.*` is probably similar as the above one, but it also caused lots of infra flakiness without having the retry logic.

The quota error as you mentioned in https://github.com/kubernetes-sigs/kubetest2/pull/13#issuecomment-668292386 actually has a different error message - `Quota exceeded for quota metric 'Requests' and limit 'Requests per minute' of service 'container.googleapis.com'`, and I definitely agree we should *NOT* retry on this error. And the version error has already been solved by https://github.com/kubernetes-sigs/kubetest2/pull/66 so we don't need retry for it either.

kubetest2-gke/deployer/up.go
chizhg: Or we can add an extra check to disallow setting `--backup-regions` and `--backup-zones` when `CI` is not true - https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables?

chizhg: Also if the two backup flags are not set, which is the default, the changes in this PR will just be a no-op. So it's not a breaking Chang for existing test flows.

kubetest2-gke/deployer/up.go
amwat: If it works after retrying in a different zone that just seems like a coincidence :) 
But that is very much a legitimate issue in most of our gke components testing, which shouldn't be retried.

pkg/exec/exec.go
amwat: not sure why we can't use it here? 
`io.Writer` is an interface both `io.MultiWriter` and `*bytes.Buffer` satisfy that interface.

kubetest2-gke/deployer/up.go
amwat: relying on janitor is not an ideal solution. janitor is only supposed to be used when say jobs are aborted. A successful kubetest2 --up --down should not rely on boskos janitor existing. 

Why is the down flow broken? if you mean that cluster delete command fails if cluster wasn't created in the first place then that's WAI?
We definitely don't take care of resources that are brought up behind the scenes during cluster creation, but if we are creating any resources/firewall rules as part of `Up` then those should be cleaned up in `Down` otherwise that's a bug.

currently as this stands if we have 2 clusters and 2 zones (1 original and 1 backup)
and cluster 1 is created in zone1 but cluster2 fails
we are retrying everything in zone2
and if both clusters come up in zone2 then cluster1 in zone1 is leaked.

chizhg: Currently for all the jobs we run with GKE clusters, we are not using the `--down` flag in order to save time in CI, so we are completely relying on janitor to do the cleanups, and we haven't seen any issues so far.

> Why is the down flow broken? if you mean that cluster delete command fails if cluster wasn't created in the first place then that's WAI?
Yes, the networks and firewall rules will be leaked if the cluster wasn't created in the first place, but we have janitor to clean them up so it's not an issue in CI.

It will complicate the logic a lot if we want to support cleaning up these leftover resources due to retrying, and also as I mentioned in another thread, this will be a no-op change if `--backup-regions` and `--backup-zones` are empty, which is by default, so it won't impact users who do not need it.

Maybe I can add description to these two flags saying `setting them will probably leak resources, please use them with caution`, and we can add a `TODO` to fix it? Or if you are completely not convinced with this change, I'll then update ASM/Istio test-infra, then we'll have this logic duplicated everywhere...

pkg/exec/exec.go
chizhg: Fixed.

kubetest2-gke/deployer/up.go
amwat: I'm definitely +1 on having a retry mode, but would like to see the suggested changes :) any new feature should be self contained and honor the full kubetest2 cluster lifecycle. We shouldn't be adding features which explicitly rely on also needing janitor.

Is there any reason why all clusters need to be recreated in a new zone, can't we just recreate the ones that fail? 
also having backup zones tells me that your tests don't really have any preference of a zone. 
Then, we can also make --zone itself take a comma separated list and choose randomly among them (and then use retries in different zones)

I believe for non stockout retriable errors, users would also be interested in retrying cluster creation even for original single zone requests.

tldr; I would love to see this being added as a more general retry feature (with all these considerations) that isn't solely for getting around stockout errors. (stockout errors can still be one of the retriable errors)

chizhg: > Is there any reason why all clusters need to be recreated in a new zone, can't we just recreate the ones that fail?

I can try that, but looks it'll make the [Up](https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gke/deployer/up.go#L37) function even more complicated, and also complicate the logic if we want to do proper cleanups since we'll need to track which regions/zones it tried to create for each cluster.

> also having backup zones tells me that your tests don't really have any preference of a zone.
Then, we can also make --zone itself take a comma separated list and choose randomly among them (and then use retries in different zones)

Currently for multicluster, we are assuming they should be created in the same region/zone, but in the future we might want them to be created in different regions/zones (the requirement was brought up by one of the engineer partners but it's not a high priority), then we'll need to change `--region` and `--zone` to take a comma separated list with its length equal to the number of clusters. That's why I'm not reusing this flag to set the backup regions/zones.

> I believe for non stockout retriable errors, users would also be interested in retrying cluster creation even for original single zone requests.

Other errors are mostly due to "the cluster was created, but the health check failed", so we will always get an error if we retry creating the cluster with **the same name** in the same region/zone. Then it'll require us deleting the clusters before recreating them, which is less efficient than directly recreating them in a different region/zone.


I can get your concern and I agree they are all valid, but I'm not sure what'll the best way to move forward without complicating the code logic here too much.

amwat: Yes, I would like to see the retry feature be discussed in more detail (especially for multi cluster) before we try to add anything, maybe as an issue/doc to cover existing+future use cases (like you mentioned). 

We can think about reworking the flags to be read from a config file if we want a lot of customizability and have each cluster's lifecycle be handled independently.

chizhg: I've created https://github.com/kubernetes-sigs/kubetest2/issues/88, please take a look when you have time. Thanks.

PR #89, opened 2021-02-02
Setup working directory correctly for bazel staging

amwat:
and various other fixes as a follow up to https://github.com/kubernetes-sigs/kubetest2/pull/80

/cc @spiffxp @BenTheElder @MushuEE 

/cc @liggitt


pkg/build/bazel.go
```diff
@@ -37,6 +37,7 @@ func (b *Bazel) Stage(version string) error {
 	location := b.StageLocation + "/v" + version
 	klog.V(0).Infof("Staging builds to %s ...", location)
 	cmd := exec.Command("bazel", "run", "//:push-build", "--", location)
+	cmd.SetDir(b.RepoRoot)
```
BenTheElder, line 40: FWIW kind had this same bug 🙃

pkg/build/options.go
```diff
@@ -36,19 +45,19 @@ func (o *Options) Validate() error {
 
 func (o *Options) implementationFromStrategy() error {
 	switch o.Strategy {
```
BenTheElder, line 47: nit: feels like we should do switch `bazelStrategy(o.Strategy)` or make the `o.Strategy` type a `bazelStrategy` and compare to the constants directly but /shrug

pkg/build/krel.go
```diff
@@ -48,6 +48,7 @@ func (rpb *ReleasePushBuild) Stage(version string) error {
 	}
 
 	// currently krel requires to be run from kubernetes root
+	// should support passing working directory
```
BenTheElder, line 51: we should probably help patch this as well, having to `os.Chdir` is brittle, particularly for "library" code.

amwat: created https://github.com/kubernetes/release/issues/1891

kubetest2-gce/deployer/build.go
```diff
@@ -39,8 +40,16 @@ func (d *deployer) Build() error {
 		if err != nil {
 			return err
 		}
+
 		// append the kubetest2 run id
-		version += "+" + d.commonOptions.RunID()
+		// avoid double + in the version
+		// so they are valid docker tags
+		if strings.Contains(version, "+") {
```
MushuEE, line 47: We have any tests for these?

pkg/build/options.go
```diff
@@ -20,6 +20,15 @@ import (
 	"fmt"
 )
 
+type buildStrategy string
```
MushuEE, line 23: Should we make the stuct `BuildAndStageStrategy and couple the Build method to the stage method so that below we can just use
```
Strategy.Stage{
			RepoRoot:      o.RepoRoot,
			StageLocation: o.StageLocation,				
			ImageLocation: o.ImageLocation,				
}
```
Or something to that effect and avoid the cases? ((not sure if this actually works because the builder and stagers are a little differnt

kubetest2-gce/deployer/build.go
amwat: we have them over on the gke deployer side, but good point we shouid have some here as well.

we lack a lot of unit tests in general majorly because most of the things are cmd.exec /shrug 
I'll be working on improving that story

pkg/build/options.go
amwat: so technically stage is optional. defaults are noops.
bazel build and staging is already coupled.

we technically shouldn't be forcing the krel stager when choosing the make builder (should leave it as noop and expose choosing the stager independently as well). 

But there's not a lot of customizability here anyway (krel stager is the only one available currently)

PR #90, opened 2021-02-16
Refactor to simplify tar extraction logic

justinsb:
Switch to a map rather than hard-coding.


amwat (commented):
/lgtm
/approve

Thanks!

PR #91, opened 2021-02-16
Refactor artifacts logic

justinsb:
In particular this makes the distinction between the artifacts dir and the
run dir (but doesn't change the logic yet), but also allows us to download
and run ginkgo without requiring it to be in the PATH (if ARTIFACTS is not
set).

Builds on #90


pkg/app/cmd.go
```diff
@@ -260,14 +233,14 @@ func (o *options) ShouldTest() bool {
 	return o.test != ""
 }
 
-func (o *options) ArtifactsDir() string {
```
amwat, line 263: we might need to expose this again in the future if this and rundir are decoupled.

pkg/app/app.go
```diff
@@ -120,7 +120,7 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		exec.InheritOutput(test)
 
 		envsForTester := os.Environ()
-		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "ARTIFACTS", opts.ArtifactsDir()))
+		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "ARTIFACTS", opts.RunDir()))
```
amwat, line 123: let's also additionally expose this as `KUBETEST2_RUN_DIR`

pkg/testers/ginkgo/package.go
```diff
@@ -72,7 +73,12 @@ func (t *Tester) AcquireTestPackage() error {
 }
 
 func (t *Tester) extractBinaries(downloadPath string) error {
-	// finally, search for the test package and extract it
+	// ensure the artifacts dir
+	if err := os.MkdirAll(artifacts.BaseDir(), os.ModePerm); err != nil {
```
amwat, line 77: I think we already do this as part of the main kubetest2 app code.

pkg/app/app.go
```diff
@@ -57,12 +57,12 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 	// TODO(bentheelder): signal handling & timeout
 
 	// ensure the artifacts dir
-	if err := os.MkdirAll(opts.ArtifactsDir(), os.ModePerm); err != nil {
+	if err := os.MkdirAll(opts.RunDir(), os.ModePerm); err != nil {
```
amwat, line 60: we ensure the directory exists, can drop from ginkgo tester.

amwat, line 60: nit: since we are already changing stuff related to this 
let's `klog.Infof(opts.RunDir())` similar to how we log the runID on lin 87 for consistency

amwat (commented):
Huge +1 to this, Thanks!

Ideally we want to expose different options, so deployers can decide for themselves which artifacts are 
1. ephemeral (e.g. os.TempDir() ), 
2. not ephemeral but not upload-worthy (RunDir()),  
3. not ephemeral, usable across runs (os.UserCacheDir() )
4. not ephemeral, upload-worthy (ArtifactsDir()).

/lgtm
/approve
/hold for a few nits

pkg/app/app.go
MushuEE: nit: Also update that comment on 59 to match*

Do we ensure it exists with correct permissions?

pkg/testers/ginkgo/package.go
```diff
@@ -82,18 +88,23 @@ func (t *Tester) extractBinaries(downloadPath string) error {
...
+		"kubernetes/test/bin/e2e.test": t.e2eTestPath,
+		"kubernetes/test/bin/ginkgo":   t.ginkgoPath,
+	}
+	extracted := map[string]bool{}
+
 	for {
 		// Put this check before any break condition so we don't
 		// accidentally incorrectly error
-		if foundTestPackage && foundBinary {
+		if len(extracted) == len(extract) {
```
MushuEE, line 107: Do you want to return here or break to also get the double check on the downloadPath check?

pkg/app/app.go
justinsb: @MushuEE Updated the comment.

@amwat I actually added it to the ginkgo tester because I hit the directory not existing, because the ginkgo tester might be invoked from e.g. [kubetest2-tester-kops](https://github.com/kubernetes/kops/blob/master/tests/e2e/kubetest2-tester-kops/main.go) and doesn't necessarily go through this path (though I'm still learning how this all fits together, so could be wrong here).

justinsb: Good idea - done!

pkg/app/app.go
justinsb: I like that - setting us up for the future cleanup (I like the 4 directories you suggested in your comment - just wanted to keep the PR focused).  Added!

pkg/app/cmd.go
justinsb: Agree, but now we should use `artifacts.BaseDir()` (note that we've moved the flag there, invoked by `artifacts.BindFlags()`)

pkg/testers/ginkgo/package.go
justinsb: Yes, but kubetest2-test-kops calls this directly

pkg/testers/ginkgo/package.go
justinsb: Good call - I think I get away with it, but a break is cleaner.  Changed!

justinsb (commented):
Thanks - pushed updates!

pkg/app/app.go
amwat: hmm if you invoke the kubetest2-tester-kops binary directly (not recommended) then you would hit this.
but if you invoke it as `kubetest2 kops --test=kops` then it should go through the kubetest2 lifecycle.

amwat: /hold cancel

interested in getting this in, can be cleaned up later.

PR #93, opened 2021-02-19
Add testsuite name in the generated junit-runner.xml file

chizhg:
Some tools would be unhappy if the testsuite in the JUnit XML files does not have a name. This PR adds `kubetest2` as the testsuite name in the generated `junit-runner.xml` file.


amwat (commented):
/lgtm
/approve

PR #94, opened 2021-02-22
[GKE] Add support for gke_make build

amwat:
/cc @liggitt @listx @MushuEE


kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -0,0 +1,108 @@
...
+	klog.V(2).Infof("starting gke build ...")
+	// Skip validation for faster builds
+	// TODO: add support for a separate validate mode
+	if err := gmb.runWithActions(os.Stdout, os.Stderr, []gkeBuildAction{compile, pack}, arg("VERSION_SUFFIX", gmb.VersionSuffix)); err != nil {
+		return "", err
+	}
+	version := &bytes.Buffer{}
+	if err := gmb.runWithActions(version, ioutil.Discard, []gkeBuildAction{printVersion}, arg("VERSION_SUFFIX", gmb.VersionSuffix)); err != nil {
+		return "", err
+	}
```
liggitt, line 86: suggest running this before compile/package so if it fails we fail fast

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -0,0 +1,108 @@
...
+	return strings.TrimSuffix(version.String(), "\n"), nil
```
liggitt, line 87: strings.TrimSpace? require the result to be non-empty?

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (gmb *GKEMake) Build() (string, error) {
+	klog.V(2).Infof("starting gke build ...")
+	// Skip validation for faster builds
+	// TODO: add support for a separate validate mode
+	if err := gmb.runWithActions(os.Stdout, os.Stderr, []gkeBuildAction{compile, pack}, arg("VERSION_SUFFIX", gmb.VersionSuffix)); err != nil {
+		return "", err
+	}
+	version := &bytes.Buffer{}
+	if err := gmb.runWithActions(version, ioutil.Discard, []gkeBuildAction{printVersion}, arg("VERSION_SUFFIX", gmb.VersionSuffix)); err != nil {
```
liggitt, line 84: capture stderr and echo if err != nil?

kubetest2-gke/deployer/build/gke_make.go
amwat: done

kubetest2-gke/deployer/build/gke_make.go
amwat: done

kubetest2-gke/deployer/build/gke_make.go
amwat: done

pkg/build/options.go
```diff
@@ -20,13 +20,14 @@ import (
...
+type BuildAndStageStrategy string //nolint:golint
 
 const (
 	// bazelStrategy builds and (optionally) stages using bazel
-	bazelStrategy buildStrategy = "bazel"
-	// makeStrategy builds using make and (optionally) stages using krel
-	makeStrategy buildStrategy = "make"
+	bazelStrategy BuildAndStageStrategy = "bazel"
+	// MakeStrategy builds using make and (optionally) stages using krel
+	MakeStrategy BuildAndStageStrategy = "make"
```
MushuEE, line 30: nit: did you mean for one to be public and the other not?

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -0,0 +1,117 @@
...
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	if err := gmb.runWithActions(stdout, stderr, []gkeBuildAction{printVersion}, arg("VERSION_SUFFIX", gmb.VersionSuffix)); err != nil {
+		klog.Errorf("failed to get version: %s\n%v", stderr.String(), err)
+		return "", err
+	}
+
+	version := strings.TrimSpace(stdout.String())
+	if version == "" {
+		return "", fmt.Errorf("failed to get version: got empty version")
```
liggitt, line 87: non-blocking: print stderr here, might be useful things in it?

liggitt (commented):
one suggestion about printing stderr, feel free to disregard. I'll defer lgtm to kubetest2 folks, but all the bits I care about look good here.

pkg/build/options.go
amwat: yes, the other one doesn't need to be exported, it isn't used outside the package

kubetest2-gke/deployer/build/gke_make.go
amwat: done

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -0,0 +1,118 @@
...
+
+	version := strings.TrimSpace(stdout.String())
+	if version == "" {
+		klog.Error(stderr.String())
+		return "", fmt.Errorf("failed to get version: got empty version")
+	}
+
+	// Skip validation for faster builds
+	// TODO: add support for a separate validate mode
+	if err := gmb.runWithActions(os.Stdout, os.Stderr, []gkeBuildAction{compile, pack}, arg("VERSION_SUFFIX", gmb.VersionSuffix)); err != nil {
```
listx, line 93: I think you can just do

```
arg("VERSION", version)
```

here (instead of passing in the VERSION_SUFFIX and having the build re-calculate to the same version string as the earlier call).

amwat: done

Issue #87, opened 2021-01-27
Support deployers passing additional information to testers


rifelpet:
Currently a tester is invoked with only arguments provided from the kubetest2 invocation (`kubetest2 ... -- --tester-args-here`) along with some environment variables. The deployer can optionally provide a kubeconfig file to the tester via an interface:

https://github.com/kubernetes-sigs/kubetest2/blob/d56fa28ed1f35903661fac5f57503e0dedf29b42/pkg/app/app.go#L128-L133

As I build kubetest2 support for Kops I'm noticing that certain testers have flags whose values are only known by the deployer rather than the process invoking kubetest2.

For example a [kubectl test](https://github.com/kubernetes/kubernetes/blob/0221f68ac39f5e83936b112f2bea8a9c3b01a9da/test/e2e/kubectl/kubectl.go#L444-L446) relies on a [Host flag](https://github.com/kubernetes/kubernetes/blob/0221f68ac39f5e83936b112f2bea8a9c3b01a9da/test/e2e/framework/test_context.go#L294), but the cluster's apiserver host name isn't known at the time that kubetest2 is invoked, so appending it to the end of the kubetest2 command isn't possible. Many of the `cloudConfig` flags in `test_context.go` have values that aren't known until the deployer has deployed the cluster too.

It would be nice to have a way for a deployer to provide a set of arguments and/or environment variables to the tester. One idea is to add a `DeployerWithTesterArgs` interface similar to the `DeployerWithKubeconfig` interface:

```go
type DeployerWithTesterArgs interface {
	Deployer

	TesterArgs() []string
}
```

One concern I have with that is the deployer would need to know which tester is being used in order to provide it the necessary arguments. The tester interface doesn't provide any sort of identifier that we could pass into the new interface's function. We could add a `Name` to `Tester`:

```go
type DeployerWithTesterArgs interface {
	Deployer

	TesterArgs(name string) []string
}

type Tester struct {
	Name string
	TesterPath string
	TesterArgs []string
}
```

but this increasingly tight coupling feels contradictory to [kubetest2's objectives](https://docs.google.com/document/d/10Zfu0KiJVMxHL-rNMW9pOESdFcdCuw2ABrDFKVQqooM/edit#heading=h.by3gwc1npkp4).

Any thoughts or ideas on how to provide testers with additional info they need from deployers? I'm happy to implement something if we reach a consensus.

amwat:
[...]

PR #95, opened 2021-02-25
Extend support for testing from locally built artifacts

amwat:
Add functionality to the ginkgo tester to be able to test from current builds.
somewhat makes it easier for https://github.com/kubernetes-sigs/kubetest2/issues/87

This will help solve skew problems e.g. kubectl https://testgrid.k8s.io/conformance-all#Conformance%20-%20GCE%20-%20master%20-%20kubetest2

Also refactor kind deployer a bit.

local testing is now as easy as :)

```
kubetest2 kind --build --up --test=ginkgo -- --focus-regex='Kubectl.client.Kubectl.replace.should.update.a.single-container'
```

/cc @BenTheElder @spiffxp

### DIFF
--- a/kubetest2-gce/deployer/build.go
+++ b/kubetest2-gce/deployer/build.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"k8s.io/klog"
+	"sigs.k8s.io/kubetest2/pkg/build"
 
 	"sigs.k8s.io/kubetest2/pkg/exec"
 )
@@ -56,6 +57,7 @@ func (d *deployer) Build() error {
 				return fmt.Errorf("error staging build: %v", err)
 			}
 		}
+		build.StoreCommonBinaries(d.RepoRoot, d.commonOptions.RunDir())
 	} else {
 		// this code path supports the kubernetes/cloud-provider-gcp build
 		klog.V(2).Info("starting the build")
--- a/kubetest2-gce/deployer/build.go
+++ b/kubetest2-gce/deployer/build.go
@@ -22,8 +22,8 @@ import (
 	"strings"
 
 	"k8s.io/klog"
-	"sigs.k8s.io/kubetest2/pkg/build"
 
+	"sigs.k8s.io/kubetest2/pkg/build"
 	"sigs.k8s.io/kubetest2/pkg/exec"
 )
 
--- a/kubetest2-gke/deployer/build.go
+++ b/kubetest2-gke/deployer/build.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"k8s.io/klog"
+
 	"sigs.k8s.io/kubetest2/pkg/build"
 )
 
--- a/kubetest2-gke/deployer/build.go
+++ b/kubetest2-gke/deployer/build.go
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"k8s.io/klog"
+	"sigs.k8s.io/kubetest2/pkg/build"
 )
 
 var (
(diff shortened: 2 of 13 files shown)


kubetest2-gce/deployer/build.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"k8s.io/klog"
+	"sigs.k8s.io/kubetest2/pkg/build"
```
BenTheElder, line 25: nit: group with import below (this is a first party import)

kubetest2-gke/deployer/build.go
```diff
@@ -23,6 +23,7 @@ import (
 	"strings"
 
 	"k8s.io/klog"
+	"sigs.k8s.io/kubetest2/pkg/build"
```
BenTheElder, line 26: same here with import grouping, first party imports go in their own group

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -135,9 +159,15 @@ func (t *Tester) Execute() error {
 		return nil
 	}
 
+	t.initKubetest2Info()
 	return t.Test()
 }
 
+// initializes relevant information from the well defined kubetest2 environment variables.
+func (t *Tester) initKubetest2Info() {
+	t.runDir = os.Getenv("KUBETEST2_RUN_DIR")
```
BenTheElder, line 168: no defaulting?

pkg/fs/fs.go
```diff
@@ -0,0 +1,61 @@
...
+		return err
+	}
+	defer in.Close()
+	// create dst file
+	// this is like f, err := os.Create(dst); os.Chmod(f.Name(), src.Mode())
+	out, err := os.OpenFile(dst, os.O_RDWR|os.O_CREATE|os.O_TRUNC, info.Mode())
+	if err != nil {
+		return err
+	}
+	// make sure we close the file
```
BenTheElder, line 47: why is this error reported but not in.Close() ?

pkg/app/app.go
```diff
@@ -124,6 +124,9 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 
 		envsForTester := os.Environ()
 		// We expose both ARIFACTS and KUBETEST2_RUN_DIR so we can more granular about caching vs output in future.
+		// also add run_dir to $PATH for locally built binaries
+		dollarPath := opts.RunDir() + string(filepath.ListSeparator) + os.Getenv("PATH")
```
BenTheElder, line 128: nit: `updatedPath`, `$` is specific to shell.

kubetest2-gke/deployer/build.go
BenTheElder: this also applies to other files

pkg/build/build.go
```diff
@@ -67,3 +72,30 @@ func sourceVersion(kubeRoot string) (string, error) {
 	}
 	return "", fmt.Errorf("could not find kubernetes version in output: %q", strings.Join(output, "\n"))
 }
+
+var (
+	CommonTestBinaries = []string{
```
spiffxp, line 77: Any reason this needs to be a `var`?
```suggestion
const (
	CommonTestBinaries = []string{
```

spiffxp (approved):
/approve
/lgtm
I have one nit but it can be done as followup

pkg/build/build.go
amwat: go doesn't support array constants? 
https://golang.org/ref/spec#Constants

PR #96, opened 2021-02-27
[GCE] Use the cluster configuration from config-test.sh

amwat:
Explicitly set `KUBE_CONFIG_FILE=config-test.sh` similar to kubetest (https://github.com/kubernetes/test-infra/blob/d18ba04e55424699d33a47a066e4883b398c2a27/kubetest/e2e.go#L63)

This is where standard configurations specifically for end-to-end test clusters using kube-up are set.

fixes: https://github.com/kubernetes/kubernetes/issues/99480

/cc @BenTheElder @liggitt

### DIFF
--- a/kubetest2-gce/deployer/common.go
+++ b/kubetest2-gce/deployer/common.go
@@ -46,14 +46,6 @@ func (d *deployer) initialize() error {
 		}
 	}
 
-	if d.commonOptions.ShouldUp() || d.commonOptions.ShouldDown() {
-		path, err := d.verifyKubectl()
-		if err != nil {
-			return err
-		}
-		d.kubectlPath = path
-	}
-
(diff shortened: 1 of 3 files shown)


kubetest2-gce/deployer/common.go
```diff
@@ -170,6 +162,16 @@ func (d *deployer) buildEnv() []string {
 	// used by the cluster/ scripts
 	env = append(env, fmt.Sprintf("KUBECTL_PATH=%s", d.kubectlPath))
 
+	// KUBE_CONFIG_FILE determines the file to use for setting up configuration for the cluster
```
BenTheElder, line 165: this name is still so unfortunate 🙃

BenTheElder (approved):
/lgtm
/approve

Issue #99, opened 2021-03-02
[GCE] [Failing Test] conformance nodePort tests


amwat:
https://testgrid.k8s.io/conformance-all#Conformance%20-%20GCE%20-%20master%20-%20kubetest2

xref: https://github.com/kubernetes/enhancements/issues/2464

After moving to `config-test.sh` in https://github.com/kubernetes-sigs/kubetest2/pull/96 the default GCE network name changed from `default` to `e2e-test-${USER}`
so changes initially added in https://github.com/kubernetes-sigs/kubetest2/pull/34 are not taking effect.

/cc @BenTheElder @spiffxp

BenTheElder:
and the gce network name breaks the test...? 😬

amwat:
The tests need extra setup to open up the nodePort range which needs to know the name of the network, we were defaulting to the wrong one.

PR #100, opened 2021-03-02
[GCE] Explicitly set KUBE_GCE_NETWORK

amwat:
fixes: https://github.com/kubernetes-sigs/kubetest2/issues/99

propagates to

https://github.com/kubernetes/kubernetes/blob/0ced9d2854174778163b791038c0a757103455b6/cluster/gce/config-test.sh#L134

Also use a run specific prefix instead of a static one to easily be able to tie back to a run.

Tested locally with

```
kubetest2 gce --gcp-project=<my_project> --gcp-zone=<my_zone> --repo-root=<repo_root> --up --test=ginkgo -- --focus-regex='Services.should.be.able.to.create.a.functioning.NodePort.service' --use-built-binaries
```

### DIFF
--- a/kubetest2-gce/deployer/common.go
+++ b/kubetest2-gce/deployer/common.go
@@ -129,9 +129,13 @@ func (d *deployer) buildEnv() []string {
 	// the run of kubetest2 and so should be placed in the artifacts directory
 	env = append(env, fmt.Sprintf("KUBECONFIG=%s", d.kubeconfigPath))
 
-	// kube-up and kube-down get this as a default ("kubernetes") but log-dump
-	// does not. opted to set it manually here for maximum consistency
+	// kube-up and kube-down get this as a default ("kubernetes" / "e2e-test-${USER}")
+	// but log-dump does not, set it explicitly here for maximum consistency
 	env = append(env, fmt.Sprintf("KUBE_GCE_INSTANCE_PREFIX=%s", d.instancePrefix))
+
+	// kube-up and kube-down get this as a default ("default" / "e2e-test-${USER}")
+	// but log-dump does not, set it explicitly here for maximum consistency
+	env = append(env, fmt.Sprintf("KUBE_GCE_NETWORK=%s", d.network))
 
 	// Pass through number of nodes and associated IP range. In the future,
 	// IP range will be configurable.
--- a/kubetest2-gce/deployer/deployer.go
+++ b/kubetest2-gce/deployer/deployer.go
@@ -81,6 +81,27 @@ type deployer struct {
 	CreateCustomNetwork         bool   `desc:"Sets the environment variable CREATE_CUSTOM_NETWORK=true during deployment."`
 }
 
+// pseudoUniqueSubstring returns a substring of a UUID
+// that can be reasonably used in resource names
+// where length is constrained
+// e.g https://cloud.google.com/compute/docs/naming-resources
+// but still retain as much uniqueness as possible
+// also easily lets us tie it back to a run
+func pseudoUniqueSubstring(uuid string) string {
+	// both KUBETEST2_RUN_ID and PROW_JOB_ID uuids are generated
+	// following RFC 4122 https://tools.ietf.org/html/rfc4122
+	// e.g. 09a2565a-7ac6-11eb-a603-2218f636630c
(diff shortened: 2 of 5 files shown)


kubetest2-gce/deployer/common.go
```diff
@@ -129,10 +129,14 @@ func (d *deployer) buildEnv() []string {
...
 	env = append(env, fmt.Sprintf("KUBECONFIG=%s", d.kubeconfigPath))
 
-	// kube-up and kube-down get this as a default ("kubernetes") but log-dump
-	// does not. opted to set it manually here for maximum consistency
+	// kube-up and kube-down get this as a default ("kubernetes" / "e2e-test-${USER}")
+	// but log-dump does not, set it explicitly here for maximum consistency
 	env = append(env, fmt.Sprintf("KUBE_GCE_INSTANCE_PREFIX=%s", d.instancePrefix))
 
+	// kube-up and kube-down get this as a default ("default" / "e2e-test-${USER}")
+	// but log-dump does not, set it explicitly here for maximum consistency
```
BenTheElder, line 137: neat. did kubetest do this? are there more of these?

kubetest2-gce/deployer/deployer.go
```diff
@@ -81,6 +81,27 @@ type deployer struct {
...
+// that can be reasonably used in resource names
+// where length is constrained
+// e.g https://cloud.google.com/compute/docs/naming-resources
+// but still retain as much uniqueness as possible
+// also easily lets us tie it back to a run
+func pseudoUniqueSubstring(uuid string) string {
+	// both KUBETEST2_RUN_ID and PROW_JOB_ID uuids are generated
+	// following RFC 4122 https://tools.ietf.org/html/rfc4122
+	// e.g. 09a2565a-7ac6-11eb-a603-2218f636630c
+	// extract the first 13 characters (09a2565a-7ac6) as they are the ones that depend on
```
BenTheElder, line 94: if we're constrained to 12/36 characters, perhaps we should skip the fixed `-`?

BenTheElder: we also probably don't want purely timestamp bits if we can avoid it, else concurrent runs can conflict ...?

amwat: I mean currently it's a fixed name for all so it's no worse

kubetest2-gce/deployer/common.go
amwat: > did kubetest do this ?

I don't think so it just relied on the defaults.

> are there more of these?

Hopefully not 😅

amwat:
@spiffxp how are we managing ACLs for the build cluster?

```
Activated service account credentials for: [pr-kubekins@kubernetes-jenkins-pull.iam.gserviceaccount.com]
```

```
ERROR: (gcloud.compute.instances.add-metadata) Could not fetch resource:
 - Required 'compute.instances.get' permission for 'projects/k8s-prow-builds/zones/us-central1-b/instances/kt2-5a1d0484-7af2-master'
```

amwat:
Anyway the actual error is

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubetest2/100/gce-build-up-down/1366553371506380800#1:build-log.txt%3A893

```
ERROR: (gcloud.compute.firewall-rules.delete) Could not fetch resource:
 - The resource 'projects/k8s-jkns-e2e-gce-slow-1-2/global/firewalls/kt2-5a1d0484-7af2-minion-nodeports' was not found
```

because it was already deleted by `kube-down`  (now that we plumb through the right network)
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubetest2/100/gce-build-up-down/1366553371506380800#1:build-log.txt%3A882

```
Deleting firewall rules remaining in network kt2-5a1d0484-7af2: kt2-5a1d0484-7af2-minion-nodeports
Deleted [https://www.googleapis.com/compute/v1/projects/k8s-jkns-e2e-gce-slow-1-2/global/firewalls/kt2-5a1d0484-7af2-minion-nodeports].
```


https://github.com/kubernetes/kubernetes/blob/6d64dfea641c8cc8ade0d5402abbc5685ba6f70c/cluster/gce/util.sh#L2473

spiffxp (approved):
/approve
/lgtm

Issue #45, opened 2020-08-17
Should have a noop deployer


amwat:
Currently kubetest2 requires a deployer entrypoint.
One common use case is to run tests on an existing cluster, for which the deployer is mostly a noop.
kubetest has something similar: https://github.com/kubernetes/test-infra/blob/master/kubetest/none.go

amwat:
FYI @pymander

randomvariable:
Interested in this as we're a potential consumer in Cluster API.

We've also inverted the deployer flow somewhat, where we have a test framework that deploys clusters using Cluster API and then runs kubetest.

amwat:
the #sig-testing channel and meeting in the kubernetes slack is the best place (https://github.com/kubernetes-sigs/kubetest2#community-discussion-contribution-and-support)

Yup, definitely should consider a kubetest2 CAP* deployer (since that would solve the exact use case you described)

randomvariable:
Thanks, will create a new issue for tracking.

PR #103, opened 2021-03-05
[Node] Add node-e2e tester

amwat:
Also add a noop deployer to go with it.

Started with `REMOTE=true` only since most testing uses that.

fixes: https://github.com/kubernetes-sigs/kubetest2/issues/45

ref:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/e2e-node-tests.md#delete-instance-after-tests-run
ref: https://github.com/kubernetes/kubernetes/blob/96be00df69390ed41b8ec22facc43bcbb9c88aae/build/root/Makefile#L206-L271
ref: https://github.com/kubernetes/kubernetes/blob/master/hack/make-rules/test-e2e-node.sh

Currently fully relies on `make test-e2e-node` which itself does a build https://github.com/kubernetes/kubernetes/blob/cea1d4e20b4a7886d8ff65f34c6d4f95efcb4742/test/e2e_node/builder/build.go, 
might be possible to breakout the build part separately into a dedicated node deployer in the future.

Also sadly, local runs don't stream the test logs everything is dumped when the tests finish

tested locally with

```
kubetest2 noop -v2 --test=node -- --repo-root=<repo-root> --gcp-zone=<zone> --gcp-project=<project> -v2 --focus-regex='should.not.show.its.pid.in.the.non-hostpid.containers.\[NodeFeature:HostAccess\]' --test-args='--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"'
```

started with porting over functionality needed by `pull-kubernetes-node-e2e`

/cc @BenTheElder @spiffxp @dims

### DIFF
--- a/kubetest2-noop/deployer/deployer.go
+++ b/kubetest2-noop/deployer/deployer.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package deployer implements the kubetest2 kind deployer
+package deployer
+
+import (
+	"flag"
+	"os"
+	"path/filepath"
+
+	"github.com/octago/sflags/gen/gpflag"
+	"github.com/spf13/pflag"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kubetest2/pkg/types"
+)
+
+// Name is the name of the deployer
+const Name = "noop"
+
+// New implements deployer.New for kind
+func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
+	// create a deployer object and set fields that are not flag controlled
+	d := &deployer{
+		commonOptions: opts,
+	}
+	// register flags and return
+	return d, bindFlags(d)
+}
+
+// assert that New implements types.NewDeployer
+var _ types.NewDeployer = New
+
+type deployer struct {
+	// generic parts
+	commonOptions types.Options
+
+	KubeconfigPath string `flag:"kubeconfig" desc:"Absolute path to existing kubeconfig for cluster"`
+}
+
+func (d *deployer) Up() error {
+	return nil
+}
+
+func (d *deployer) Down() error {
+	return nil
+}
+
+func (d *deployer) IsUp() (up bool, err error) {
+	return false, nil
+}
+
+func (d *deployer) DumpClusterLogs() error {
+	return nil
+}
+
+func (d *deployer) Build() error {
+	// TODO: build should probably still exist with common options
+	return nil
+}
+
+func (d *deployer) Kubeconfig() (string, error) {
+	// noop deployer is specifically used with an existing cluster and KUBECONFIG
+	if d.KubeconfigPath != "" {
+		return d.KubeconfigPath, nil
+	}
+	if kconfig, ok := os.LookupEnv("KUBECONFIG"); ok {
+		return kconfig, nil
+	}
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(home, ".kube", "config"), nil
+}
+
+// helper used to create & bind a flagset to the deployer
+func bindFlags(d *deployer) *pflag.FlagSet {
+	flags, err := gpflag.Parse(d)
+	if err != nil {
+		klog.Fatalf("unable to generate flags from deployer")
+		return nil
+	}
+
+	klog.InitFlags(nil)
+	flags.AddGoFlagSet(flag.CommandLine)
+
+	return flags
+}
+
+// assert that deployer implements types.DeployerWithKubeconfig
+var _ types.DeployerWithKubeconfig = &deployer{}
--- a/pkg/testers/node/node.go
+++ b/pkg/testers/node/node.go
@@ -1,0 +1,179 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package node implements a node tester that implements e2e node testing following
+// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-node/e2e-node-tests.md#delete-instance-after-tests-run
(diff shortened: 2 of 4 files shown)


amwat:
FYI @SergeyKanzhelev

amwat:
ping

ehashman:
What is this?

BenTheElder:
kubetest2 replaces https://github.com/kubernetes/test-infra/tree/master/kubetest (deprecated).
this PR introduces support for invoking the node_e2e test suite

kubetest2-noop/deployer/deployer.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/octago/sflags/gen/gpflag"
+	"github.com/spf13/pflag"
+	"k8s.io/klog"
+
+	"sigs.k8s.io/kubetest2/pkg/types"
+)
+
+// Name is the name of the deployer
+const Name = "noop"
```
BenTheElder, line 33: this should probably be a seperate PR, so folks from SIG node can focus on the e2e_node specific parts.

pkg/testers/node/node.go
```diff
@@ -0,0 +1,179 @@
...
+	"sigs.k8s.io/kubetest2/pkg/boskos"
+	"sigs.k8s.io/kubetest2/pkg/exec"
+)
+
+const (
+	target                 = "test-e2e-node"
+	gceProjectResourceType = "gce-project"
+)
+
+type Tester struct {
```
bobbypage, line 42: how do we pass in `--image-config-file`? https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/runner/remote/run_remote.go#L55

Many tests pass in custom image file, e.g. https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/kubernetes/test-infra%24+file:%5Econfig/jobs+--image-config-file%3D&patternType=literal

pkg/testers/node/node.go
```diff
@@ -0,0 +1,179 @@
...
+func (t *Tester) constructArgs() []string {
+	defaultArgs := []string{
+		"REMOTE=true",
+		"DELETE_INSTANCES=true",
+	}
+
+	argsFromFlags := []string{
+		"SKIP=" + t.SkipRegex,
+		"FOCUS=" + t.FocusRegex,
+		"RUNTIME=" + t.Runtime,
```
bobbypage, line 154: I'm not sure if just exposing `RUNTIME` is enough. For example, take this job: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-node/crio.yaml#L52

```
--node-test-args=--container-runtime=remote --container-runtime-endpoint=unix:///var/run/crio/crio.sock ...
```

I'm not clear how that translates into equivalent args here...

kubetest2-noop/deployer/deployer.go
amwat: Ack

pkg/testers/node/node.go
amwat: So this PR isn't the full migration, I anticipate we'll port over specific functionality as and when needed. Infact I'd prefer that so we don't just blindly port over all the flags.

I chose the subset of things used by `pull-kubernetes-node-e2e` as a start

amwat: Note that power users can still set the corrsponding ENVs before invoking kubetest2 (until kubetest2 supports them as flags (if at all))

pkg/testers/node/node.go
amwat: Note that power users can still set the corrsponding ENVs before invoking kubetest2 (until kubetest2 supports them as flags (if at all))

pkg/testers/node/node.go
bobbypage: Got it, makes sense, would be good clarify that in the PR description that goal here is just `pull-kubernetes-node-e2e` for now.

pkg/testers/node/node.go
bobbypage: > Note that power users can still set the corrsponding ENVs before invoking kubetest2

How does that work? Doesn't the make cmd execution need to inherit the bash env variables, i.e. something like `cmd.Env = os.Environ()`.

Additionally, I think the challenge is it's not clear what exact envs you need set because the current jobs using kubetest(1) node e2e runner took in more or less same args as [run_remote.go](https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/runner/remote/run_remote.go). However since this is based on `test-e2e-node.sh`, for all those jobs, you'll need to translate those args to the equivalent envs from `test-e2e-node` which isn't super straightforward.

amwat: https://golang.org/pkg/os/exec/#Cmd

cmd will inherit all exported envs from parent process by default.

> I think the challenge is it's not clear what exact envs you need set

Ack, but I think that is orthogonal to the kubetest2 tester invocation. i.e. they come more from https://github.com/kubernetes/kubernetes/blob/96be00df69390ed41b8ec22facc43bcbb9c88aae/build/root/Makefile#L206-L271 

Ideally, we will port over all relevant functionality as flags over time.

I think the main question at hand is whether or not `sig-node` thinks `make test-e2e-node` is a better common entrypoint for node testing as opposed what we have currently https://github.com/kubernetes/test-infra/blob/4be647a393b6b491a2c0f00145d8bb9f8416b082/kubetest/e2e.go#L556-L611

pkg/testers/node/node.go
amwat: Ack, updated PR description

pkg/testers/node/node.go
bobbypage: ack got it.

> I think the main question at hand is whether or not sig-node thinks make test-e2e-node is a better common entrypoint for node testing as opposed what we have currently

Yup, this is the main question. 

As I mentioned the earlier, the problem I see is if the goal is for us is to migrate all e2e node jobs to kubetest2 eventually this will basically require rewriting all of the job configs since the current kubetest1 interface doesn't rely on `make test-e2e-node`. On other side standardizing on `make test-e2e-node` maybe a good move eventually :) but worth to discuss.   

Folks in the [sig-node-ci group](https://docs.google.com/document/d/1fb-ugvgdSVIkkuJ388_nhp2pBTy_4HEVg5848Xy7n5U) would be candidates to ask about this. 

cc a couple folks to get their thoughts on usage of `make test-e2e-node` vs existing kubetest node e2e interface.

/cc @SergeyKanzhelev @ehashman @fromanirh @knabben

harche: +1 for standardizing on `make test-e2e-node`

BenTheElder:
FYI kubetest is definitely deprecated, so we would still appreciate eyes on the migration, given node e2e is a bit different from the swaths of cluster e2e jobs SIG Testing tends to be most familiar with.

spiffxp:
/assign @ehashman @SergeyKanzhelev @derekwaynecarr 
Can y'all find someone from SIG Node who is willing to take a look at this? The goal is to make running node-e2e less painful, and a thing that node owns

dims:
Ran this:
```
kubetest2 noop -v2 --test=node -- --repo-root=/home/dims/go/src/k8s.io/kubernetes --gcp-zone=$MY_ZONE --gcp-project=$MY_PROJECT -v2 --focus-regex='should.not.show.its.pid.in.the.non-hostpid.containers.\[NodeFeature:HostAccess\]' --test-args='--kubelet-flags="--cgroups-per-qos=true --cgroup-root=/"'
```

got this:
```
Ran 1 of 354 Specs in 142.957 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 353 Skipped


Ginkgo ran 1 suite in 2m24.56251924s
Test Suite Passed

Success Finished Test Suite on Host test-cos-beta-89-16108-403-11
```

There are definitely some rough edges around credentials, ssh keys, etc that we will eventually need to work through. Let's merge and iterate!

dims:
/approve
/lgtm

thanks @amwat

Issue #102, opened 2021-03-05
Support Version Marker Publishing


rifelpet:
I'm migrating a few remaining kOps e2e jobs to kubetest2 and I'm facing a kubetest2 limitation that I'd like to address upstream. We have a pipeline of jobs using kubetest 1 that publishes version markers upon tests passing. These markers indicate a version of kOps to use in other kubetest2 jobs.

I'm hoping to have kubetest2-kops publish version markers but I'm facing the following challenges:

* The Deployer interface doesn't know whether the tests passed or failed. The deployer knows the exact version to publish in a version marker, but it can't determine whether or not to publish the marker.
* The Tester interface doesn't know the version to publish as a marker.

This likely relates to my https://github.com/kubernetes-sigs/kubetest2/issues/87 but I'm wondering if it would simpler for the Deployer to know whether the tester succeeded or failed. We could add an extension like the other [DeployerWith](https://github.com/kubernetes-sigs/kubetest2/blob/b6e52f6d52afc2f2977de4fa7498883b73bd9713/pkg/types/types.go#L83-L102) interfaces, perhaps something like:

```go
type DeployerWithTesterResults interface {
	OnSuccess() error
	OnFailure() error
}
```

I'm happy to open a PR if an implementation is agreed upon.

For reference, [here](https://github.com/kubernetes/test-infra/blob/0c6e1a7eab573f5fb8f48f14c6c87248640d6c43/config/jobs/kubernetes/kops/kops-pipeline.yaml#L46) is one of the remaining kOps jobs that uses [this](https://github.com/kubernetes/test-infra/blob/0c6e1a7eab573f5fb8f48f14c6c87248640d6c43/kubetest/kops.go#L782-L797) kubetest 1 functionality.

/cc @justinsb

amwat:
I think this proposal sounds very reasonable :) since this is very much in line with the whole lifecycle.
#87 was a bit awkward since we ideally want to plug and play deployers and testers, so deployer knowing TestArgs is an anti-pattern we'd like to avoid.

but PostTest() actions sounds like a good addition.
maybe instead of OnSuccess() and OnFailure()
just a

```go
type DeployerWithPostTester interface {
           PostTest(testError error) error
}

```

FYI something like this is also good for say DumpingClusterLogs after tests.

which can be invoked after test() if it exists.


Also, a big thanks for all the migration efforts for kops kubetest2 :) 

xref: https://github.com/kubernetes/enhancements/issues/2464

PR #104, opened 2021-03-05
Add DeployerWithPostTester interface

rifelpet:
This allows the deployer to define after-test behavior based on the results of the test.

Unfortunately PostTest can't be wrapped in writer.WrapStep because PostTest has an error parameter but WrapStep expects `func() error`.

If both test.Run and PostTest return errors, PostTest's error is returned.

Closes https://github.com/kubernetes-sigs/kubetest2/issues/102


amwat (commented):
/approve
/lgtm

Thanks!

hopefully this will open up a lot of custom use cases!

amwat (commented):
/lgtm

Issue #58, opened 2020-09-21
The error message in the JUnit xml result is not informative


chizhg:
When we use `kubetest2` to run a test workflow, for each phase, it will add a test in the JUnit xml result if it fails, and include the error message, the main reason for adding this is for integration with Testgrid and Spyglass.

However, the error message for the test is something like `exit status 255` and does not contain any information in detail, for example https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-client-continuous/1308147648712151041.

Two issues related to this:
1. Other phases like `Up` and `Down` also have this issue, it would be nice if some detailed information can be added in the error message.
2. Since most of the test flows will generate the JUnit xml files themselves (lots of the projects use `gotestsum`), is adding the `Test` phase into the JUnit result a bit noisy? (It is already causing a bit confusion around the Knative community).

/kind bug

amwat:
Now that we've split the testers into separate binaries, we can probably skip wrapping the Test phase into a Junit error (but that's only for use cases where the tester itself is generating it's own Junit.) For simple cases especially with the `--test=exec` most commands you'll run would be simple commands that won't generate Junit so we still need an overall pass/fail for the entire Test phase.


For up/down we definitely do need to improve the output there, we should be using `metadata.NewJUnitError` in the deployers in most places.

PR #105, opened 2021-03-08
Add classname attr to the JUnit test case and add a flag to allow skipping test junit report

chizhg:
1. Add `classname` attribute to the JUnit test case to show on Spyglass, this will fix https://github.com/kubernetes-sigs/kubetest2/issues/107
2. For some tester implementations, they will be generating their own Junit results, in which cases the extra `Test` JUnit result will be redundant. So add an extra flag `--skip-test-junit-report` to allow skipping reporting the JUnit test result. This will partially fix https://github.com/kubernetes-sigs/kubetest2/issues/58.

### DIFF
--- a/pkg/app/cmd.go
+++ b/pkg/app/cmd.go
@@ -184,12 +184,13 @@ func splitArgs(args []string) ([]string, []string) {
 
 // options holds flag values and implements deployer.Options
 type options struct {
-	help  bool
-	build bool
-	up    bool
-	down  bool
-	test  string
-	runid string
+	help                bool
+	build               bool
+	up                  bool
+	down                bool
+	test                string
+	skipTestJUnitReport bool
+	runid               string
 }
 
 // bindFlags registers all first class kubetest2 flags
@@ -199,6 +200,8 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.up, "up", false, "provision the test cluster")
 	flags.BoolVar(&o.down, "down", false, "tear down the test cluster")
 	flags.StringVar(&o.test, "test", "", "test type to run, if unset no tests will run")
+	flags.BoolVar(&o.skipTestJUnitReport, "skip-test-junit-report", false, "skip reporting the test step as a JUnit test case, "+
+		"should be set to true when solely relying on the tester binary to generate it's own junit.")
 
 	var defaultRunID string
 	// reuse uid for CI use cases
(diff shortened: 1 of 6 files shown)


amwat (commented):
I wouldn't want to mess with the name too much as it invalidates a lot of testgrid history.

ref: https://github.com/kubernetes/test-infra/pull/21180#issuecomment-793063005

I think filtering out the kubetest2 specific entries can be done by whatever is post processing it by many other means? (sorting and relying on capitalized alphabetical order is not the most ideal)

chizhg:
> I wouldn't want to mess with the name too much as it invalidates a lot of testgrid history.
> 
> ref: [kubernetes/test-infra#21180 (comment)](https://github.com/kubernetes/test-infra/pull/21180#issuecomment-793063005)
> 
> I think filtering out the kubetest2 specific entries can be done by whatever is post processing it by many other means? (sorting and relying on capitalized alphabetical order is not the most ideal)

That makes total sense. I've reverted the change in my PR and updated the commit message accordingly.

pkg/app/cmd.go
```diff
@@ -199,6 +200,7 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 	flags.BoolVar(&o.up, "up", false, "provision the test cluster")
 	flags.BoolVar(&o.down, "down", false, "tear down the test cluster")
 	flags.StringVar(&o.test, "test", "", "test type to run, if unset no tests will run")
+	flags.BoolVar(&o.skipTestJUnitReport, "skip-test-junit-report", false, "skip reporting the test result as a JUnit test case")
```
spiffxp, line 203: Is "test" a necessary distinction here? 

Vs "skip-junit-report", "skip reporting kubetest2 steps to junit.xml", etc

spiffxp (changes requested):

pkg/app/cmd.go
amwat: IIUC the intention is to only skip adding the junit entry for the `kubetest2.Test()` step. build, up, down will still show.
It's not too ideal but since we have explicit tester binaries I can see use cases where it should be treated specially.

amwat, line 203:
```suggestion
	flags.BoolVar(&o.skipTestJUnitReport, "skip-test-junit-report", false, "skip reporting the test step as a JUnit test case. used when solely relying on the tester binary to generate it's own junit.")
```

amwat (commented):
/approve

pkg/app/cmd.go
chizhg: Done.

chizhg: Thanks @amwat . Yes, that is the intention for this change.

spiffxp (approved):
/lgtm

PR #111, opened 2021-03-12
Kubetest2 Version Addition

### LINKED ISSUE #108: Add a version command
`kubetest2 --version` should output the tag/commit that it's built from.

This is especially important since currently kubetest2 is used in a lot of place by go getting (mostly latest)
so knowing which exact binary(commit) was used is useful. 
This will be further useful if it gets surfaced in testgrid metadata for any kubetest2 based job.

Additionally, this should be done for all the in-tree deployers and testers.

The version will also reflect the release once we move to tagged releases.
related: https://github.com/kubernetes-sigs/kubetest2/issues/17

A typical useful pattern is to inject and set package variables at build time through linker flags.
https://blog.alexellis.io/inject-build-time-vars-golang/

A good example can be found in the kind repo

https://github.com/kubernetes-sigs/kind/blob/fbf6cdcf5240c6b994657c9a4f85657ee0591331/Makefile#L57

https://github.com/kubernetes-sigs/kind/blob/fbf6cdcf5240c6b994657c9a4f85657ee0591331/pkg/cmd/kind/version/version.go

/help
/good-first-issue

supriya-premkumar:
- Uses ldflags to capture HEAD rev in short form.
- Prints built commit version by default
- Adds --version support
- Fixes partially #108

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,9 @@
 # get the repo root and output path
 REPO_ROOT:=$(shell pwd)
 export REPO_ROOT
-OUT_DIR?=$(REPO_ROOT)/bin
+OUT_DIR=$(REPO_ROOT)/bin
+# record the source commit in the binary, overridable
+COMMIT?=$(shell git describe --tags --always --dirty 2>/dev/null)
 INSTALL?=install
 # make install will place binaries here
 # the default path attempts to mimic go install
@@ -47,7 +49,7 @@ SPACE:=$(subst ,, )
 SHELL:=env PATH=$(subst $(SPACE),\$(SPACE),$(PATH)) $(SHELL)
 # ==============================================================================
 # flags for reproducible go builds
-BUILD_FLAGS?=-trimpath -ldflags="-buildid="
+BUILD_FLAGS?=-trimpath -ldflags="-buildid= -X=sigs.k8s.io/kubetest2/pkg/app.GitTag=$(COMMIT)"
 
 build-all:
 	go build -v $(BUILD_FLAGS) ./...
--- a/pkg/app/cmd.go
+++ b/pkg/app/cmd.go
@@ -67,6 +67,8 @@ func runE(
 	opts.bindFlags(kubetest2Flags)
 	artifacts.MustBindFlags(kubetest2Flags)
 
+	cmd.Printf("Running deployer %s version: %s\n", deployerName, shim.GitTag)
+
 	// NOTE: unknown flags are forwarded to the deployer as arguments
 	kubetest2Flags.ParseErrorsWhitelist.UnknownFlags = true
 
@@ -91,6 +93,7 @@ func runE(
 		if err != nil {
 			return fmt.Errorf("unable to find tester %v: %v", opts.test, err)
 		}
+		cmd.Printf("Running tester %s version: %s\n", opts.test, shim.GitTag)
 
 		// Get tester usage by running it with --help
 		var helpArgs []string
--- a/pkg/app/shim/shim.go
+++ b/pkg/app/shim/shim.go
@@ -26,6 +26,9 @@ import (
 	"sigs.k8s.io/kubetest2/pkg/process"
 )
 
+// GitTag captures the git commit SHA of the build. This gets printed by all the deployers and testers
+var GitTag string
+
 // Main implements the kubetest2 root binary entrypoint
 func Main() {
 	if err := Run(); err != nil {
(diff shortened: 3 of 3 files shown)


spiffxp (commented):
/lgtm
Code looks sound

/hold
Deferring to @amwat for approval

Looks like this gets version in the log. What needs to happen to get it in metadata.json and picked up by kettle?

Do we need multiple versions for different kubetest2 binaries? Deployer at one version tester at another sort of thing?

amwat (commented):
/approve

Thank you for working on this!

lgtm after 2 suggestions

Also, we want to do this for all the deployer and tester binaries as well, but fine if it's not in this PR (in which case let's modify the PR description so that it won't close the issue.)

amwat:
> Looks like this gets version in the log. What needs to happen to get it in metadata.json and picked up by kettle?

We don't produce a `metadata.json` file yet. Should be fairly straightforward to generate one with the version populated under `RunDir()`.

Likely a good new functionality addition under `pkg/app/metadata`.

Fine if it's a follow up and left out from this PR.

supriya-premkumar:
> all the deployer and tester binaries as well,


> Also, we want to do this for all the deployer and tester binaries as well, but fine if it's not in this PR

I didn't know that every deployer gets built as a separate binary. Let me fix that.

Makefile
```diff
@@ -21,6 +21,8 @@
 REPO_ROOT:=$(shell pwd)
 export REPO_ROOT
 OUT_DIR=$(REPO_ROOT)/bin
+# record the source commit in the binary, overridable
+COMMIT?=$(shell git rev-parse --short HEAD 2>/dev/null)
```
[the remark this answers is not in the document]
supriya-premkumar: @amwat Just want to make sure that you won't see a `-dirty` suffix if there are any untracked files.
`-dirty` works only on tracked diffs. Is this the expectation?

pkg/app/cmd.go
```diff
@@ -67,6 +67,9 @@ func runE(
 	opts.bindFlags(kubetest2Flags)
 	artifacts.MustBindFlags(kubetest2Flags)
 
+	if opts.test == "" {
+		cmd.Printf("Running deployer %s version: %s\n", deployerName, shim.GitTag)
```
amwat, line 71: we can log deployer version always, no need to guard it only when `opts.test==""` 
deployer always runs, tester only runs when `opts.test !=""`

Makefile
```diff
@@ -47,7 +49,7 @@ SPACE:=$(subst ,, )
 SHELL:=env PATH=$(subst $(SPACE),\$(SPACE),$(PATH)) $(SHELL)
 # ==============================================================================
 # flags for reproducible go builds
-BUILD_FLAGS?=-trimpath -ldflags="-buildid="
+BUILD_FLAGS?=-trimpath -ldflags="-buildid= -X=sigs.k8s.io/kubetest2/pkg/app.GitTag=$(COMMIT)"
```
amwat, line 52: unfortunately, handling the deployers won't be so straightforward :) 
this will just always get the kubetest2 shim version. but we can have deployers built at different commits.

amwat (commented):
Handling deployers will be tricky since we want to *support* out of tree deployers/testers as well (but we'll be adding the functionality only to the in-tree ones).

I'm fine leaving out the deployer versions from this PR, can be done as a follow up.

What we'd want to achieve that is:

For deployers:

- add `BUILD_FLAGS` specifically in the `install-deployer-%` make rule
- populate a `var` at build time like we already are doing.
- would be good to follow the same convention, say `var GitTag string` under respective `main.go` 
- plumb this through by adding a new callback (for backwards compatibility) `func MainWithVersion(deployerName string, deployerVersion string, newDeployer types.NewDeployer)`.
- modify `Run` to accept version (will be passed as empty from `Main`) `func Run(deployerName string, deployerVersion string, newDeployer types.NewDeployer) error` -> `runE(cmd, args, deployerName, deployerVersion, newDeployer)`

For testers:

- add `BUILD_FLAGS` specifically in the `install-tester-%` make rule
- populate a `var` at build time like we already are doing.
- would be good to follow the same convention, say `var GitTag string` under respective `main.go` 
- invoke the tester with `--version` similar to how we handle `--help` for testers (but don't fail on errors) https://github.com/kubernetes-sigs/kubetest2/blob/master/pkg/app/cmd.go#L95-L105

Makefile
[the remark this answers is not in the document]
amwat: if we can add that as well, that would be great :)  but not a huge blocker.

maybe with `git status --porcelain` and manually add a `-dirty` suffix if we find untracked files?

supriya-premkumar:
I was held up last week. I will work on it this week, thank you.

pkg/app/cmd.go
```diff
@@ -67,6 +67,8 @@ func runE(
 	opts.bindFlags(kubetest2Flags)
 	artifacts.MustBindFlags(kubetest2Flags)
 
+	cmd.Printf("Running deployer %s version: %s\n", deployerName, shim.GitTag)
```
amwat, line 70: we can add this after deployer versions are added

pkg/app/cmd.go
```diff
@@ -91,6 +93,7 @@ func runE(
 		if err != nil {
 			return fmt.Errorf("unable to find tester %v: %v", opts.test, err)
 		}
+		cmd.Printf("Running tester %s version: %s\n", opts.test, shim.GitTag)
```
amwat, line 96: we can add this after testers versions are added

amwat (commented):
getting the version for the main binary is a good start

/lgtm
/approve
/hold

modifying the PR description so it doesn't close the issue

BenTheElder:
we might be able to get clever with https://golang.org/pkg/runtime/debug/#BuildInfo for out of tree stuff importing the repo

pkg/app/shim/shim.go
```diff
@@ -66,22 +69,29 @@ func NewCommand() *cobra.Command {
...
 func runE(cmd *cobra.Command, args []string) error {
+	cmd.Printf("Running %s version: %s\n", BinaryName, GitTag)
 	// there should be at least one argument (the deployer) unless the user
 	// is asking for help on the shim itself
 	if len(args) < 1 {
 		return cmd.Help()
 	}
 
-	// gracefully handle -h or --help if it is the only argument
+	// gracefully handle help or version command if it is the only argument
```
MushuEE, line 79: Unrelated to this feature, but if other arguments are set AND --help is included, it will try to deploy?

amwat: no https://github.com/kubernetes-sigs/kubetest2/blob/96ab8c9c5ca183b362a8f95d24bd328e2527ab70/pkg/app/cmd.go#L139-L143

BenTheElder:
@amwat perhaps unhold and iterate?

amwat:
yeah, I think we can iterate on this

/unhold

PR #115, opened 2021-03-20
For private clusters, do not use --create-subnetwork if it's multi-project profile

chizhg:
Since for shared-vpc - the network topology we use for multi-project profile, the subnetworks will be created before cluster creation, so for private cluster + shared vpc, if we also specify `--create-subnetwork`, we will get an error which says `Cannot specify both --subnetwork and --create-subnetwork at the same time.`

This PR updates to the logic to only add `--create-subnetwork` if it's single-project profile.


amwat (commented):
/lgtm
/approve

Issue #109, opened 2021-03-11
[GCE] support reusing existing ssh keys


amwat:
/cc @spiffxp 
xref: https://github.com/kubernetes/k8s.io/issues/1751#issuecomment-790171952

xref: https://github.com/kubernetes/enhancements/issues/2464

amwat:
I'd like to deprecate "JENKINS" while doing this 

https://github.com/kubernetes/test-infra/blob/31c626f6a2e660c8cf8d44b417cb5a71d0d63dba/config/prow/config.yaml#L639-L647

best way would be to export the same files as both `JENKINS_GCE_SSH_PRIVATE_KEY_FILE` ,`GCE_SSH_PRIVATE_KEY_FILE` and `JENKINS_GCE_SSH_PUBLIC_KEY_FILE`,  `GCE_SSH_PUBLIC_KEY_FILE`

as part of `preset-k8s-ssh` so that both kubetest and kubetest2 jobs can use the same preset.

kubetest2 will rely only on `GCE_SSH_PRIVATE_KEY_FILE` and `GCE_SSH_PUBLIC_KEY_FILE`

thoughts @spiffxp ?

spiffxp:
SGTM

PR #117, opened 2021-03-23
[GCE] Allow reusing ssh keys in CI

amwat:
fixes: #109 

also flip unnecessarily blocking similar gke check to default not block.

/cc @spiffxp @BenTheElder


spiffxp (approved):
I will try to submit my bits in the form of a followup PR. Would like to confirm this works first

spiffxp (approved):
/approve
/lgtm

PR #124, opened 2021-03-29
Shorten PULL_BASE_SHA

amwat:
few more fixes for https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kubetest2-push-binaries/1376655559062720512

/cc @BenTheElder


hack/ci/push-binaries/push-binaries.sh
```diff
@@ -21,24 +21,29 @@ set -o xtrace
...
 # pass through git details from prow / image builder
 if [ -n "${PULL_BASE_SHA:-}" ]; then
   export COMMIT="${PULL_BASE_SHA:?}"
 else
-  COMMIT="$(git rev-parse --short HEAD 2>/dev/null)"
+  COMMIT="$(git rev-parse HEAD 2>/dev/null)"
   export COMMIT
 fi
 
+# short commit is currently 8 characters
```
BenTheElder, line 36: fun fact this can change as the repo gets more commits
you may want to consider a fixed size. IIRC kubernetes ran into this

PR #127, opened 2021-04-05
Let CI override $USER

amwat:
/cc @MushuEE @spiffxp 

Currently failing to dump logs
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-kubernetes-gce-conformance-latest-kubetest2/1379089716992806912

ref: https://github.com/kubernetes-sigs/kubetest2/issues/109

Event though uid is actually 0(root) in the container we want to use whatever prow sets as long as it's not root.


MushuEE (commented):
/lgtm

PR #129, opened 2021-04-16
gke deployer supports creating GKE Autopilot clusters

chizhg:
gke deployer supports creating GKE Autopilot clusters.

GKE Autopilot creation command only supports a subset of args of the general cluster creation command, see https://cloud.google.com/sdk/gcloud/reference/container/clusters/create-auto.

### DIFF
--- a/kubetest2-gke/deployer/up.go
+++ b/kubetest2-gke/deployer/up.go
@@ -67,22 +67,26 @@ func (d *deployer) Up() error {
 	loc := location(d.region, d.zone)
 	for i := range d.projects {
 		project := d.projects[i]
-		subNetworkArgs := subNetworkArgs(d.projects, d.region, d.network, i)
+		subNetworkArgs := subNetworkArgs(d.autopilot, d.projects, d.region, d.network, i)
 		for j := range d.projectClustersLayout[project] {
 			cluster := d.projectClustersLayout[project][j]
 			privateClusterArgs := privateClusterArgs(d.projects, d.network, d.privateClusterAccessLevel, d.privateClusterMasterIPRanges, cluster)
 			eg.Go(func() error {
 				// Create the cluster
-				args := make([]string, len(d.createCommand()))
-				copy(args, d.createCommand())
+				args := d.createCommand()
 				args = append(args,
 					"--project="+project,
 					loc,
-					"--machine-type="+d.machineType,
-					"--image-type="+image,
-					"--num-nodes="+strconv.Itoa(d.nodes),
 					"--network="+transformNetworkName(d.projects, d.network),
 				)
+				// A few args are not supported in GKE Autopilot cluster creation, so they should be left unset.
+				// https://cloud.google.com/sdk/gcloud/reference/container/clusters/create-auto
+				if !d.autopilot {
+					args = append(args, "--machine-type="+d.machineType)
+					args = append(args, "--num-nodes="+strconv.Itoa(d.nodes))
+					args = append(args, "--image-type="+d.imageType)
+				}
+
 				if d.workloadIdentityEnabled {
 					args = append(args, fmt.Sprintf("--workload-pool=%s.svc.id.goog", project))
 				}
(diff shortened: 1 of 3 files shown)


kubetest2-gke/deployer/up.go
```diff
@@ -78,11 +78,21 @@ func (d *deployer) Up() error {
...
+				// when they are not provided via the flags.
+				// https://cloud.google.com/sdk/gcloud/reference/container/clusters/create-auto
+				if d.machineType != "" {
+					args = append(args, "--machine-type="+d.machineType)
+				}
+				if d.nodes != 0 {
+					args = append(args, "--num-nodes="+strconv.Itoa(d.nodes))
+				}
+				if d.imageType != "" {
+					args = append(args, "--image-type="+image)
```
zkoopmans, line 93: This is a bug: 

args = append(args, "--image-type="d.imageType)

chizhg: Fixed.

kubetest2-gke/deployer/up.go
```diff
@@ -78,11 +78,21 @@ func (d *deployer) Up() error {
...
+				if d.machineType != "" {
+					args = append(args, "--machine-type="+d.machineType)
+				}
+				if d.nodes != 0 {
+					args = append(args, "--num-nodes="+strconv.Itoa(d.nodes))
+				}
+				if d.imageType != "" {
+					args = append(args, "--image-type="+d.imageType)
+				}
+
```
amwat, line 95: IIUC the current changes only fixes assumptions while generating the flags , but users still need to use the power "create-command" override?

I think we should just simply support autopilot first class, by adding an auto pilot option?

`kubetest2 gke --up --autopilot`

 and move this validation/generation to `verifyUpFlags()`

chizhg: I considered about this option before making the change. The question is how we should handle the case when the user provides both `--autopilot` and `--create-command=container clusters create xxx yyy` flags. If we throw an error, that is even worse than simply relying on "create-command" override. If we replace `create` with `create-auto`, that could confuse users..

I think the best way to support this is breaking down `--create-command` into three flags:
`--gcloud-command-version`: can be alpha, beta, or default
`--autopilot`: can be true or false
`--gcloud-extra-flags`: can be a string of extra flags we want to add to the gcloud command
But since kubetest2 is not versioned now, we'll probably need to consider backward compatibility when we make the change.

WDYT? Any better suggestions?

amwat: **Overall I'd want to see a state where if `create-command` is specified everything else should be ignored** (well super ideally this doesn't need to exist, but it's almost impossible to cover all cases so we keep it for power users) .

Relying on a mix of both cluster creation flags of the kubetest2 tool and the custom create command definitely is not great.

> --gcloud-command-version: can be alpha, beta, or default
> --autopilot: can be true or false
> --gcloud-extra-flags: can be a string of extra flags we want to add to the gcloud command

This approach sounds reasonable to me. I don't think we need to worry about backwards compatibility for new flags? (we'll still keep the `create-command` flag around)

chizhg: They need to be mutual exclusive since `--create-command` can be configured to include all the 3 new flags, so I added a logic to ignore `--create-command` if any of these 3 new flags are explicitly specified. Please let me know if it looks good to you.

amwat: can we invert it? let's treat `--create-command` as an explicit override and ignore anything else, and can update the flag help text accordingly.

chizhg: Done.

kubetest2-gke/deployer/up.go
```diff
@@ -67,22 +67,26 @@ func (d *deployer) Up() error {
...
-					"--num-nodes="+strconv.Itoa(d.nodes),
 					"--network="+transformNetworkName(d.projects, d.network),
 				)
+				// A few args are not supported in GKE Autopilot cluster creation, so they should be left unset.
+				// https://cloud.google.com/sdk/gcloud/reference/container/clusters/create-auto
+				if !d.autopilot {
+					args = append(args, "--machine-type="+d.machineType)
+					args = append(args, "--num-nodes="+strconv.Itoa(d.nodes))
+					args = append(args, "--image-type="+d.imageType)
+				}
```
amwat, line 88: It might be worth refactoring all the argument builder logic somewhere as a follow up.

amwat (commented):
/lgtm
/approve

PR #135, opened 2021-05-24
Change the functions for kubetest2-gke deployer to public

chizhg:
Change the functions for kubetest2-gke deployer to public, so that it can be used as a library.

### DIFF
--- a/kubetest2-gke/deployer/options/up.go
+++ b/kubetest2-gke/deployer/options/up.go
@@ -19,7 +19,30 @@ package options
 import "fmt"
 
 type UpOptions struct {
-	NumClusters int `flag:"~num-clusters" desc:"Number of clusters to create, will auto-generate names as (kt2-<run-id>-<index>)"`
+	GcloudCommandGroup string `flag:"~gcloud-command-group" desc:"gcloud command group, can be one of empty, alpha, beta."`
+	Autopilot          bool   `flag:"~autopilot" desc:"Whether to create GKE Autopilot clusters or not."`
+	GcloudExtraFlags   string `flag:"~gcloud-extra-flags" desc:"Extra gcloud flags to pass when creating the clusters."`
+	CreateCommandFlag  string `flag:"~create-command" desc:"gcloud subcommand and additional flags used to create a cluster, such as container clusters create --quiet. If it's specified, --gcloud-command-group, --autopilot, --gcloud-extra-flags will be ignored."`
+
+	NumClusters             int    `flag:"~num-clusters" desc:"Number of clusters to create, will auto-generate names as (kt2-<run-id>-<index>)."`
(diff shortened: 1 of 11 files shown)


kubetest2-gke/deployer/options/up.go
```diff
@@ -19,7 +19,28 @@ package options
...
+
+	PrivateClusterAccessLevel    string   `flag:"~private-cluster-access-level" desc:"Private cluster access level, if not empty, must be one of 'no', 'limited' or 'unrestricted'. See the details in https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters."`
+	PrivateClusterMasterIPRanges []string `flag:"~private-cluster-master-ip-range" desc:"Private cluster master IP ranges. It should be IPv4 CIDR(s), and its length must be the same as the number of clusters if private cluster is requested."`
+	SubnetworkRanges             []string `flag:"~subnetwork-ranges" desc:"Subnetwork ranges as required for shared VPC setup as described in https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#creating_a_network_and_two_subnets. For multi-project profile, it is required and should be in the format of 10.0.4.0/22 10.0.32.0/20 10.4.0.0/14,172.16.4.0/22 172.16.16.0/20 172.16.4.0/22, where the subnetworks configuration for different project are separated by comma, and the ranges of each subnetwork configuration is separated by space."`
```
amwat, line 43: Thank you for this long overdue refactoring!

amwat (commented):
/lgtm

one quick question: there's probably more value in actually exporting the *non-deployer* "helper" functions (i.e. network setup etc.)

Usability of deployers as libraries is questionable considering the contract between deployers and testers is still binary invocation.

kubetest2-gke/deployer/options/up.go
chizhg: yw!

chizhg:
> /lgtm
> 
> one quick question: there's probably more value in actually exporting the _non-deployer_ "helper" functions (i.e. network setup etc.)
> 
> Usability of deployers as libraries is questionable considering the contract between deployers and testers is still binary invocation.

The major/only reason we want this change is to decouple the different steps (project setup, network setup, cluster creation, etc.) in the pipeline, so that we can run each step individually.

We are not planning to rebuild a new pipeline by calling these public functions, that'll be still orchestrated by the kubetest2 binary.

Does it answer your question?

amwat:
My main question is around how we plan to use the public deployer functions. If users still needs to construct the kubetest2 gke deployer then it's probably less ideal than exporting the networking pieces directly.

But either way, this will help in making these methods reusable by users who want to manage the network lifecycle on their own.

/lgtm
/approve

PR #139, opened 2021-06-10
kubetest2 gke deployer supports retries with different subnet ranges to avoid conflicts

joshua-bone:
kubetest2 gke deployer supports retries with different subnet ranges to avoid conflicts

### DIFF
--- a/kubetest2-gke/deployer/network.go
+++ b/kubetest2-gke/deployer/network.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


chizhg:
To add more context, the reason we need this is if the provided parameters have subnetwork ranges, when we retry creating the subnetworks with the same IP ranges, `gcloud` will give an `IP range conflict` error. 

This PR requires the users to provide non-overlapping subnetwork ranges for these retry operations, so that the subnetworks can be always created without getting conflicts.

amwat (commented):
Users needing to provide ranges for clusters*retryCount is a bit much.

Can we handle retries on subnets better instead?
Seems like we shouldn't need to create/delete subnets if the region remains same?
And/or creation of subnet should wait for deletion of previous subnet.

And I'm also guessing the IP conflict would only occur in the same region.

chizhg:
It's a bit more complicated than it was expected. 

1. IP conflict will occur as long as the subnets are in the same network, even if they are in the different regions.
2. The subnet cannot be deleted before the associated cluster is deleted (`gcloud` will say it's already being used and cannot be deleted). So if we want to wait for the deletion of the previous subnets, we'll have to wait for the deletion of the previous clusters as well, which means changing https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gke/deployer/up.go#L104-L109 to be blocking. This'll be adding much extra overhead to the cluster creation retries.

I definitely agree that asking users to provide ranges for clusters*retryCount is suboptimal, but that'll be a one-off config, and can reduce the turnaround time for all cluster creation retries given the `gcloud` constaints.

kubetest2-gke/deployer/network_test.go
```diff
@@ -131,3 +131,26 @@ func TestPrivateClusterArgs(t *testing.T) {
 		})
 	}
 }
+
+func TestAssertNoOverlaps(t *testing.T) {
+	testCases := []struct {
+		ranges     []string
+		shouldPass bool
+	}{
+		{[]string{"10.0.0.0/24", "11.0.0.0/24"}, true},
```
chizhg, line 140: It's better to use named fields instead of anonymous ones here.

joshua-bone: Done

kubetest2-gke/deployer/network.go
```diff
@@ -38,41 +39,115 @@ etag: %s
...
 
+func areOverlapping(r1, r2 *net.IPNet) bool {
+	return r1.Contains(r2.IP) || r2.Contains(r1.IP)
+}
+
+func (d *Deployer) internalizeNetworkFlags(numProjects int) error {
+	d.subnetworkRangesInternal = make([][]string, d.totalTryCount)
+	for tc := 0; tc < d.totalTryCount; tc++ {
+		d.subnetworkRangesInternal[tc] = make([]string, numProjects-1)
+		for p := 0; p < numProjects-1; p++ {
```
amwat, line 132: This n-1 indexing for hostProject v/s service projects is starting to become magical.

I would suggest breaking down the logic into a `SubNetwork` struct which would store the project info in a follow up.

amwat (commented):
/lgtm
/approve

/hold
one nit and a follow up suggestion. Feel free to unhold.

PR #140, opened 2021-06-14
Do not send heartbeat requests to Boskos if the interval is set to 0

chizhg:
In some cases we want to always explicitly release Boskos projects, then the heartbeat requests are not needed.

This PR changes the logic to not send heartbeat if the interval is configured as 0.


spiffxp (approved):
/approve
/lgtm
FYI @amwat

PR #141, opened 2021-06-16
Support Windows node pool in kubetest2-gke

anfernee:
When --enable-windows is specified, it will add 2 Windows node pools
into the cluster, one using LTSC image, the other using SAC image.

New options include:
--enable-windows: Whether enable Windows node pool in the cluster or
not.
--windows-num-nodes: For use with gcloud commands to specify the number
of nodes for Windows node pools in the cluster.
--windows-machine-type: For use with gcloud commands to specify the
machine type for Windows node in the cluster.

### DIFF
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -39,8 +39,9 @@ import (
 const Name = "gke"
 
 const (
-	e2eAllow     = "tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767"
-	defaultImage = "cos"
+	e2eAllow            = "tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767"
+	defaultImage        = "cos"
+	defaultWindowsImage = WindowsImageTypeLTSC
 )
 
 const (
@@ -65,9 +66,15 @@ var (
 	// m[3]: unique hash (used as nonce for firewall rules)
 	poolRe = regexp.MustCompile(`zones/([^/]+)/instanceGroupManagers/(gk[e|3]-.*-([0-9a-f]{8})-grp)$`)
 
-	urlRe           = regexp.MustCompile(`https://.*/`)
+	urlRe = regexp.MustCompile(`https://.*/`)
+
 	defaultNodePool = gkeNodePool{
 		Nodes:       3,
+		MachineType: "n1-standard-2",
+	}
+
+	defaultWindowsNodePool = gkeNodePool{
+		Nodes:       1,
 		MachineType: "n1-standard-2",
 	}
 )
@@ -154,10 +161,13 @@ func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 			Environment: "prod",
 		},
 		UpOptions: &options.UpOptions{
-			NumClusters: 1,
-			NumNodes:    defaultNodePool.Nodes,
-			MachineType: defaultNodePool.MachineType,
-			ImageType:   defaultImage,
+			NumClusters:        1,
+			NumNodes:           defaultNodePool.Nodes,
+			MachineType:        defaultNodePool.MachineType,
+			ImageType:          defaultImage,
+			WindowsNumNodes:    defaultWindowsNodePool.Nodes,
+			WindowsMachineType: defaultWindowsNodePool.MachineType,
+			WindowsImageType:   defaultWindowsImage,
 			// Leave Version as empty to use the default cluster version.
 			Version:          "",
 			GCPSSHKeyIgnored: true,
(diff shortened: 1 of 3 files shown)


kubetest2-gke/deployer/deployer.go
```diff
@@ -65,11 +65,17 @@ var (
 	// m[3]: unique hash (used as nonce for firewall rules)
 	poolRe = regexp.MustCompile(`zones/([^/]+)/instanceGroupManagers/(gk[e|3]-.*-([0-9a-f]{8})-grp)$`)
 
-	urlRe           = regexp.MustCompile(`https://.*/`)
+	urlRe = regexp.MustCompile(`https://.*/`)
```
BenTheElder, line 68: @amwat are we not linting gofmt in CI?

amwat: we should be https://github.com/kubernetes-sigs/kubetest2/blob/master/hack/tools/.golangci.yml#L21

kubetest2-gke/deployer/options/up.go
```diff
@@ -32,6 +32,9 @@ type UpOptions struct {
 	Version                 string `desc:"Use a specific GKE version e.g. 1.16.13.gke-400, 'latest' or ''. If --build is specified it will default to building kubernetes from source."`
 	WorkloadIdentityEnabled bool   `flag:"~enable-workload-identity" desc:"Whether enable workload identity for the cluster or not. See the details in https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity."`
 	GCPSSHKeyIgnored        bool   `flag:"~ignore-gcp-ssh-key" desc:"Whether the GCP SSH key should be ignored or not for bringing up the cluster."`
+	WindowsEnabled          bool   `flag:"~enable-windows" desc:"Whether enable Windows node pool in the cluster or not."`
+	WindowsNumNodes         int    `flag:"~windows-num-nodes" desc:"For use with gcloud commands to specify the number of nodes for Windows node pools in the cluster."`
+	WindowsMachineType      string `flag:"~windows-machine-type" desc:"For use with gcloud commands to specify the machine type for Windows node in the cluster."`
```
amwat, line 37: Do we need the new `NumNodes` and `MachineType` ? the existing flags should be reusable here.

amwat (commented):
Overall question:

Can't we just accomplish this by passing `--image-type=WINDOWS_LTSC` ?

The only new thing I see here is that we seems to want to create node-pools for both image types (total 2x the num_nodes). which sounds like we should be making --image-type a comma separated list instead? `--image-type=WINDOWS_LTSC,WINDOWS_SAC`

kubetest2-gke/deployer/deployer.go
anfernee: Oh, we do gofmt. I added an empty line between regexps and the default node pools. that's why it got reformatted.

kubetest2-gke/deployer/options/up.go
anfernee: Windows machine type has special requirements. Not all machine types are valid for windows. On the UI, it will force you to change the machine type, so I make it separate. 

For number of nodes, it's also for windows node pools. It takes more time for windows to come up. Normally I only create one node.

anfernee:
cc @jeremyje @liggitt

anfernee:
> Overall question:
> 
> Can't we just accomplish this by passing `--image-type=WINDOWS_LTSC` ?
> 
> The only new thing I see here is that we seems to want to create node-pools for both image types (total 2x the num_nodes). which sounds like we should be making --image-type a comma separated list instead? `--image-type=WINDOWS_LTSC,WINDOWS_SAC`

I think I could. However, the logic there could be complicated and it would affect the normal Linux workflow, which I don't want to destabilize. For example, I would have to deal with the case where linux image type is missing, it's in the middle, it's in the end. It could be really messy.

kubetest2-gke/deployer/deployer.go
BenTheElder: ah, thanks :-)

kubetest2-gke/deployer/options/up.go
BenTheElder: do you usually have windows and linux nodes? of only one node pool type each?

anfernee: Linux nodepool is required to create windows nodepool, because some cluster-wide service only runs on Linux (like kube-dns). Basically, you cannot create a windows-node-only cluster.

anfernee:
@amwat a friendly ping..

amwat (commented):
/lgtm
/approve

Thanks for clarifying, so this adds the ability to add windows node-pools to a (linux originally) cluster during bring up which lgtm.

/hold
one question:

kubetest2-gke/deployer/up.go
```diff
@@ -173,6 +181,21 @@ func (d *Deployer) CreateCluster(project string, cluster cluster, subNetworkArgs
 		//parse output for match with regex error
 		return fmt.Errorf("error creating cluster: %v, output: %q", err, output)
 	}
+
+	if d.WindowsEnabled {
```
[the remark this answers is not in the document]
jeremyje: +1 here. Having both node pools complicates things because of Windows container compatibility as well. Using this will mean that you'll always need to specify which Windows version you want to schedule on.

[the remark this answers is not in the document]
anfernee: I see. should we only add one type of node pool at one time for one cluster, and create 2 flows internally?

anfernee:
@amwat I applied the suggested change, but lost the lgtm label.

amwat:
/lgtm

Feel free to unhold after you test this out locally end-to-end.

anfernee:
can someone help remove `do-not-merge/hold`?

BenTheElder:
anyone can do that, including you 🙃  https://prow.k8s.io/command-help#hold
/hold cancel

PR #142, opened 2021-06-18
[kubetest2-gke deployer] Automatically resolves the release channel if it's not specified

chizhg:
Sometimes when the users create the GKE clusters, they just want to specify a valid cluster version but do not want to hard code the channel, but if the version is only valid in a specific channel (e.g. rapid), the cluster creation will fail.

This PR adds the support to automatically resolve the release channel if it's not specified, so that as long as the specified version is in any of the valid release channels, `kubetest2 gke --up` will work.

/cc @amwat

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/sync v0.0.0-20201207232520-09787c993a3a
 	golang.org/x/sys v0.0.0-20201221093633-bc327ba9c2f0 // indirect
+	google.golang.org/api v0.36.0
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/klog v1.0.0
 	k8s.io/release v0.7.1-0.20210204090829-09fb5e3883b8
--- a/kubetest2-gke/deployer/up.go
+++ b/kubetest2-gke/deployer/up.go
@@ -22,7 +22,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -164,6 +163,12 @@ func (d *Deployer) CreateCluster(project string, cluster cluster, subNetworkArgs
 		}
 	} else {
 		args = append(args, "--cluster-version="+d.Version)
+		releaseChannel, err := resolveReleaseChannelForClusterVersion(d.Version, locationArg)
+		if err != nil {
+			klog.Warningf("error resolving the release channel for %q: %v, will proceed with no channel", d.Version, err)
+		} else {
+			args = append(args, "--release-channel="+releaseChannel)
+		}
 	}
 	args = append(args, subNetworkArgs...)
 	args = append(args, privateClusterArgs...)
@@ -307,6 +312,9 @@ func (d *Deployer) VerifyUpFlags() error {
 	if err := validateVersion(d.Version); err != nil {
 		return err
 	}
+	if err := validateReleaseChannel(d.ReleaseChannel); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -13,7 +13,9 @@ require (
 	github.com/spf13/pflag v1.0.5
 	golang.org/x/sync v0.0.0-20201207232520-09787c993a3a
 	golang.org/x/sys v0.0.0-20201221093633-bc327ba9c2f0 // indirect
+	google.golang.org/api v0.36.0
 	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/apimachinery v0.18.8
```
amwat, line 18: Let's avoid adding k8s.io dependencies as much as possible, especially since we need it only seem to be adding this for using sets.

kubetest2-gke/deployer/version.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+		if !re.MatchString(version) {
+			return fmt.Errorf("unknown version %q", version)
+		}
+	}
+	return nil
+}
+
+func validateReleaseChannel(releaseChannel string) error {
+	if releaseChannel != "" && !validReleaseChannels.Has(releaseChannel) {
```
amwat, line 56: maybe can reuse `toReleaseChannel` to avoid needing sets.

kubetest2-gke/deployer/up.go
```diff
@@ -164,6 +163,11 @@ func (d *Deployer) CreateCluster(project string, cluster cluster, subNetworkArgs
 		}
 	} else {
 		args = append(args, "--cluster-version="+d.Version)
+		releaseChannel, err := resolveReleaseChannelForClusterVersion(d.Version, locationArg)
+		if err != nil {
+			return err
```
amwat, line 168: /hold

This should be a warning, since we also may want to test bleeding edge versions which aren't in any channel yet.

amwat (commented):
Let's make this best-effort instead of a run failing error.
Otherwise lgtm.

Isn't None channel usually a superset of all channels except maybe Rapid?

go.mod
chizhg: Done.

kubetest2-gke/deployer/version.go
chizhg: `toReleaseChannel` is used to convert the channel name in the server config to the channel name in the `gcloud` flag, so it cannot be used here. I have changed to simple array iteration to avoid using sets though.

kubetest2-gke/deployer/up.go
chizhg: Done.

kubetest2-gke/deployer/version_test.go
```diff
@@ -0,0 +1,178 @@
...
+		{
+			desc:          "the same cluster version is a match",
+			version:       "1.19.10-gke.1000",
+			target:        "1.19.10-gke.1000",
+			expectMatched: true,
+		},
+		{
+			desc:          "the same major.minor version is a match",
+			version:       "1.19",
+			target:        "1.19.10-gke.1000",
```
amwat, line 143: nit:

also add a case with `version` and `target` swapped to exercise 
```
if len(parts) < len(matchParts) {
		return false
	}
```

amwat (commented):
/lgtm
/approve

/hold

couple of nits feel free to cancel the hold.

kubetest2-gke/deployer/version.go
amwat: I meant since the strings are the same, we can use `toLower`/`toUpper` and the default case to also use it as a validation method.
But considering it's just 4 values the array solution is fine as well.

kubetest2-gke/deployer/version_test.go
chizhg: 👍 I'll address this comment in my next PR.

PR #143, opened 2021-06-18
[kubetest2-gke deployer] Normalize kubeconfig files so that they can be used everywhere without gcloud

chizhg:
When we use `gcloud clusters get-credentials` command to get the cluster credentials, the generated kubeconfig file is gcloud-dependent, e.g. the user that uses this kubeconfig file must has `gcloud` installed in the same location. This makes it hard to use the credentials in a different machine with or without `gcloud` installed.

This PR leverages the `gencred` tool in https://github.com/kubernetes/test-infra/tree/master/gencred to convert the kubeconfig files to be `gcloud` agnostic, so that they are be used everywhere.


chizhg:
This PR does not necessarily have to be in kubetest2, since I can actually do the conversion in the callsite, and it adds `k8s.io/client-go` dependency which I don't like either. Closing this PR for now.

amwat:
+1 client-go dependency is notoriously hard to maintain.

That being said, if we can get this functionality without it, kubetest2 wouldn't be a bad place for it.
kubetest2 standardizes on Kubeconfig as the contract between deployer and tester. Making the GKE kubeconfig more portable would be a good addition.

Issue #101, opened 2021-03-03
set SOURCE_DATE_EPOCH from repo when building


BenTheElder:
for reproducible builds https://github.com/kubernetes/test-infra/issues/18706#issuecomment-720947821

kubetest2 should do this internally based on the repo root.

BenTheElder:
cc @liggitt

BenTheElder:
This should be a pretty good starter issue actually: 

You'll want to integrate against https://github.com/kubernetes-sigs/kubetest2/tree/master/pkg/build

Somewhere early on, when the app has the `--build` mode enabled at least, we should identify the source location in the same way as the build logic, and compute SOURCE_DATE_EPOCH from git if possible. There is some pointer to computing that in the linked thread, basically it's just the commit timestamp as a unix epoch timestamp, set to a semi-standard environment variable (`SOURCE_DATE_EPOCH`) for reproducible builds.

BenTheElder:
Thanks again!

PR #145, opened 2021-06-24
set SOURCE_DATE_EPOCH from kubeRepo when building

xinydev:
Fixes https://github.com/kubernetes-sigs/kubetest2/issues/101. 

Set the commit timestamp of kubeRoot for SOURCE_DATE_EPOCH env for reproducibility befor bazel & make build

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -99,3 +99,17 @@ func StoreCommonBinaries(kuberoot string, outroot string) {
 		}
 	}
 }
+
+// setReproducibilityEnv set the commit timestamp of kubeRoot for SOURCE_DATE_EPOCH env for reproducibility
+func setReproducibilityEnv(kubeRoot string) {
+	if os.Getenv("SOURCE_DATE_EPOCH") != "" {
+		return
+	}
+	cmd := exec.Command("git", "log", "-1", "--pretty=%ct")
+	cmd.SetDir(kubeRoot)
+	if output, err := exec.CombinedOutputLines(cmd); err != nil {
+		klog.Warningf("failed to compute SOURCE_DATE_EPOCH from git: %v", err)
+	} else {
+		os.Setenv("SOURCE_DATE_EPOCH", output[0])
+	}
+}
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -100,7 +100,9 @@ func StoreCommonBinaries(kuberoot string, outroot string) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


BenTheElder:
cc @amwat 

I wonder if maybe we should only set the env on the build command(s) directly, so as to avoid affecting a secondary build / repo used for testing

BenTheElder:
Thanks for working on this!

pkg/build/build.go
```diff
@@ -99,3 +99,17 @@ func StoreCommonBinaries(kuberoot string, outroot string) {
 		}
 	}
 }
+
+// setReproducibilityEnv set the commit timestamp of kubeRoot for SOURCE_DATE_EPOCH env for reproducibility
```
amwat, line 103: Let's also point to the kubernetes documentation for this

```suggestion
// setReproducibilityEnv sets the SOURCE_DATE_EPOCH env to the commit timestamp of the latest commit in the
// kubernetes repository, specified under kubeRoot, for reproducible builds
// https://github.com/kubernetes/kubernetes/blob/7eae33cb0e1ead51c80ad517bc670113d77fa28d/build/README.md#reproducibility
```

pkg/build/build.go
```diff
@@ -99,3 +99,17 @@ func StoreCommonBinaries(kuberoot string, outroot string) {
...
+func setReproducibilityEnv(kubeRoot string) {
```
amwat, line 104: Let's make the method name verbose so it's clear what it's doing, since it's a known concept https://reproducible-builds.org/specs/source-date-epoch/
maybe `setSourceDateEpoch`?

pkg/build/make.go
```diff
@@ -38,6 +38,7 @@ func (m *MakeBuilder) Build() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to get version: %v", err)
 	}
+	setReproducibilityEnv(m.RepoRoot)
 	cmd := exec.Command("make", target)
```
amwat, line 42: +1 to Ben's suggestion
Let's set the env only for the specific cmd so we don't affect other things.

amwat (commented):
/approve

Thank you for working on this!
couple of minor suggestions before lgtm:

xinydev:
@amwat  Thanks a lot for the review ! PTAL

BenTheElder (approved):
/lgtm
/hold
holding in case I missed something from amit's nits, this looks good, thank you!!

amwat:
/lgtm

Thank you!

/unhold

PR #146, opened 2021-06-25
Fix to kubetest2 gke deployer supports retries with different subnet ranges to avoid conflicts

chizhg:
1. Fix to the issue in https://github.com/kubernetes-sigs/kubetest2/commit/66afb490ec30c4f12f621a63f1c6cad42247a35f#r52588671
2. Also fix one issue in parsing the server config - yaml unmarshal has some issues with parsing the nested structs, switching to json instead.

I have tested it in our existing CI flows and everything works fine.

### DIFF
--- a/kubetest2-gke/deployer/network_test.go
+++ b/kubetest2-gke/deployer/network_test.go
@@ -131,3 +131,35 @@ func TestPrivateClusterArgs(t *testing.T) {
 		})
 	}
 }
+
+func TestAssertNoOverlaps(t *testing.T) {
+	testCases := []struct {
+		ranges     []string
+		shouldPass bool
+	}{
+		{
+			ranges:     []string{"10.0.0.0/24", "11.0.0.0/24"},
+			shouldPass: true,
+		},
+		{
+			ranges:     []string{"10.0.0.0/24", "10.0.0.0/25"},
+			shouldPass: false,
+		},
+		{
+			ranges:     []string{"10.0.0.0/24", "11.0.0.0/24", "10.0.0.0/25"},
+			shouldPass: false,
+		},
+	}
+
+	for _, tc := range testCases {
+		tc := tc
+		err := assertNoOverlaps(tc.ranges)
+		if (err == nil) != tc.shouldPass {
+			if tc.shouldPass {
+				t.Errorf("test case should have passed, but failed: %q (error: %w)", tc.ranges, err)
+			} else {
+				t.Errorf("test case should have failed, but passed: %q", tc.ranges)
+			}
+		}
+	}
+}
--- a/kubetest2-gke/deployer/network_test.go
+++ b/kubetest2-gke/deployer/network_test.go
@@ -33,24 +33,6 @@ func TestPrivateClusterArgs(t *testing.T) {
 		expected       []string
 	}{
 		{
-			desc:           "no private cluster args are needed for non-private clusters",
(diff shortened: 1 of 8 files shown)


kubetest2-gke/deployer/up.go
```diff
@@ -129,7 +129,10 @@ func (d *Deployer) isRetryableError(err error) bool {
 }
 
 func (d *Deployer) CreateCluster(project string, cluster cluster, subNetworkArgs []string, locationArg string) error {
-	privateClusterArgs := privateClusterArgs(d.Projects, d.Network, d.PrivateClusterAccessLevel, d.PrivateClusterMasterIPRanges, cluster)
+	privateClusterArgs := []string{}
```
chizhg, line 132: Here is the fix.

kubetest2-gke/deployer/version.go
```diff
@@ -158,14 +158,14 @@ func isClusterVersionMatch(match, version string) bool {
...
+	out, err := exec.Output(exec.RawCommand((fmt.Sprintf("gcloud container get-server-config --format=json %s", loc))))
 	if err != nil {
 		return nil, err
 	}
 
-	// Parse the YAML into a struct.
+	// Parse the JSON into a struct.
 	cfg := &container.ServerConfig{}
-	if err := yaml.Unmarshal(out, cfg); err != nil {
+	if err := json.Unmarshal(out, cfg); err != nil {
```
chizhg, line 168: Here is the fix for parsing server config.

kubetest2-gke/deployer/up.go
amwat: nit: Can you break it up in 2 commits.
first which reverts the revert and a second which makes the fixes would be easier to track.

amwat (commented):
Thanks for quickly fixing this.

We really need end-to-end tests running upstream to catch these.
Were you able to test this out locally?

kubetest2-gke/deployer/up.go
chizhg: Sure. DOne.

test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
amwat: nit: stray file in diff

kubetest2-gke/deployer/network_test.go
```diff
@@ -32,24 +32,6 @@ func TestPrivateClusterArgs(t *testing.T) {
...
-		},
-		{
-			desc:           "--private-cluster-master-ip-range can be empty for non-private clusters",
-			projects:       []string{"project1"},
-			network:        "whatever-network",
-			accessLevel:    "",
-			masterIPRanges: []string{},
-			clusterInfo:    cluster{index: 0, name: "whatever-cluster-name"},
-			expected:       []string{},
-		},
```
amwat, line 52: why are we removing some tests?

amwat (commented):
/approve

kubetest2-gke/deployer/network_test.go
chizhg: It's because I removed the `if accessLevel == ""` check in this function, so these tests become invalid.

test.sh
chizhg: Fixed.

amwat (commented):
/lgtm

PR #147, opened 2021-06-25
Fix two bugs for kubetest2 gke retry

chizhg:
1. Return the `gcloud` output if the cluster creation command fails. This will be used for matching the retryable error pattern to determine whether to retry or not.
2. Fix the private cluster master IP ranges when retrying.

I have manually tested the change and everything works.

/cc @amwat @BenTheElder


kubetest2-gke/deployer/commandutils.go
```diff
@@ -140,7 +140,7 @@ func runWithOutputAndReturn(cmd exec.Cmd) (string, error) {
 
 	exec.SetOutput(cmd, io.MultiWriter(os.Stdout, &buf), io.MultiWriter(os.Stderr, &buf))
 	if err := cmd.Run(); err != nil {
-		return "", err
```
BenTheElder, line 143: FWIW this is possible to unit test because of our exec shim, I don't want to block fixing this feature, but it might be worth adding a quick set of tests

Issue #77, opened 2020-12-01
kubetest2 should handle signals


ixdy:
When a test job times out, Prow sends a signal to the test process to allow it to clean up before it's terminated.

Ideally, kubetest2 would handle this signal, canceling any currently-running subprocess and running any cleanup that's necessary (e.g. tearing down resources if `--down` was provided or returning resources to boskos), but it [appears this is not yet implemented](https://github.com/kubernetes-sigs/kubetest2/blob/42a5361331002149abfd284a2542fe2cef6602cf/pkg/app/app.go#L57).

ShwethaKumbla:
I am interested to contribute.

/assign

amwat:
Thanks for taking this on!

/remove-help
Please ping with any questions if you need help :)

ShwethaKumbla:
@amwat Could you suggest me some pointers to test this scenario ?

amwat:
Definitely! 

general pointers: https://gobyexample.com/signals

Overall we want to do something similar to what kubetest1 does e.g:

https://github.com/kubernetes/test-infra/blob/51fd28eca92982e2945d42001d6328158950df42/kubetest/main.go#L393

and add it to our `RealMain` flow https://github.com/kubernetes-sigs/kubetest2/blob/master/pkg/app/app.go#L48:6
most importantly we want to always call `Down` if we interrupt during `Up`.

we'll also want a wrapper on `exec` commands  (especially since the tester is invoked as a binary) similar to 
https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/kubetest/process/process.go#L142

(note all of this need not be done in a single PR)

For testing this locally, something like using the kind deployer with say the exec tester doing sleep might be easy to do locally.

`kubetest2 kind --test=exec -- sleep 100`

and then we can test interrupting this both during deploy and during test.

ShwethaKumbla:
Thanks @amwat for the pointers.

PR #148, opened 2021-06-26
Clean up resource if any signal interruption

ShwethaKumbla:
Fix for #77

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -19,7 +19,9 @@ package app
 import (
 	"fmt"
 	"os"
+	"os/signal"
 	"path/filepath"
+	"syscall"
 
 	"github.com/pkg/errors"
 	"k8s.io/klog"
@@ -75,6 +77,30 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		return errors.Wrap(err, "could not create runner output")
 	}
 	writer := metadata.NewWriter("kubetest2", junitRunner)
+
+	done := make(chan bool)
+	defer func() { done <- true }()
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+
+		// catch interrupt signals and gracefully attempt to clean up
+		for {
+			select {
+			case <-c:
+				if opts.ShouldUp() || opts.ShouldTest() {
+					klog.Info("Captured ^C, gracefully attempting to cleanup resources..")
+					if err := writer.WrapStep("Down", d.Down); err != nil {
+						result = err
+					}
+					os.Exit(0)
+				}
+			case <-done:
+				return
+			}
+		}
+	}()
+
 	// defer writing out the metadata on exit
 	// NOTE: defer is LIFO, so this should actually be the finish time
 	defer func() {
@@ -100,17 +126,6 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 			return err
 		}
 	}
-
-	// ensure tearing down the cluster happens last, even if up or test fails.
-	defer func() {
-		if opts.ShouldDown() {
-			// TODO(bentheelder): instead of keeping the first error, consider
-			// a multi-error type
-			if err := writer.WrapStep("Down", d.Down); err != nil && result == nil {
-				result = err
-			}
-		}
-	}()
 
 	// up a cluster
 	if opts.ShouldUp() {
@@ -161,6 +176,16 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 			return testErr
 		}
 	}
+
+	// ensure tearing down the cluster happens last.
+	if opts.ShouldDown() {
+		// TODO(bentheelder): instead of keeping the first error, consider
+		// a multi-error type
+		if err := writer.WrapStep("Down", d.Down); err != nil && result == nil {
+			result = err
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


ShwethaKumbla:
@amwat the fix calls Down() on signal interruption and tries to delete the created cluster. But before finishing the tasks the process getting exited. Debugging it but not getting exactly why its getting exited. I may need your help here.

pkg/app/app.go
```diff
@@ -75,6 +77,23 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
+
+	go func() {
+		// catch interrupt signals and gracefully attempt to clean up
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+
+		for range c {
+			if opts.ShouldUp() || opts.ShouldTest() {
+				klog.Info("Captured ^C, gracefully attempting to cleanup resources..")
+				if err := writer.WrapStep("Down", d.Down); err != nil {
```
amwat, line 89: So currently we already have `down()` already as a defer so I imagine before this down() gets a chance to complete the deferred down() will be also invoked and the process would exit.

We can either fully rely on the deferred Down() and just `return` here or duplicate the existing Down() logic here and remove the defer.

chizhg:
@ShwethaKumbla are you still working on this PR? It'll be great if this issue can be addressed.

ShwethaKumbla:
@chizhg Yes. Updated the PR with the latest changes. If any suggestion/feedback kindly let me know.

pkg/app/app.go
```diff
@@ -75,6 +78,33 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		return errors.Wrap(err, "could not create runner output")
 	}
 	writer := metadata.NewWriter("kubetest2", junitRunner)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
```
amwat, line 83: Do we need a wait group?

In most happy paths this signal handling shouldn't be required and won't be triggered.
We already have the done channel, so we probably don't need an additional wait group.

pkg/app/app.go
```diff
@@ -75,6 +78,33 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
+	done := make(chan bool)
```
amwat, line 84: we can defer a done here to avoid needing to do it explicitly in every exit case

`defer func(){ done <- true }()`

pkg/app/app.go
```diff
@@ -154,14 +186,16 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 
 		if dWithPostTester, ok := d.(types.DeployerWithPostTester); ok {
 			if err := dWithPostTester.PostTest(testErr); err != nil {
-				return err
+				result = err
```
amwat, line 189: can return directly if we have the `defer done`

pkg/app/app.go
```diff
@@ -97,27 +127,29 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
 
 	// ensure tearing down the cluster happens last, even if up or test fails.
-	defer func() {
-		if opts.ShouldDown() {
-			// TODO(bentheelder): instead of keeping the first error, consider
-			// a multi-error type
-			if err := writer.WrapStep("Down", d.Down); err != nil && result == nil {
-				result = err
-			}
+	if opts.ShouldDown() {
```
amwat, line 136: Now that we are removing the defer, Down() should be last after Up() and Test()

amwat (commented):
We can probably also add a few unit tests to test this functionality, but since there aren't existing tests for this file, not a blocker, fine as a follow up.

pkg/app/app.go
ShwethaKumbla: Agree with you. We do not need waitgroup since we have done channel  🤦‍♀️

pkg/app/app.go
ShwethaKumbla: If I remove defer then we need to move the condition to the end of the function, otherwise it will try to down the cluster first and then will create the cluster. So keeping defer as it is.

pkg/app/app.go
```diff
@@ -75,6 +77,34 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
+	// wg.Add(1)
+	done := make(chan bool)
+	defer func() { done <- true }()
+	go func() {
+		//defer wg.Done()
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+
+		// catch interrupt signals and gracefully attempt to clean up
+		klog.Info("waiting for interruption signal...")
```
amwat, line 91: We probably don't want to log this.

amwat (commented):
we might still need to remove the defer down as discussed earlier otherwise it will run Down() twice

ShwethaKumbla:
@amwat have squashed the commits too.

pkg/app/app.go
```diff
@@ -75,6 +77,33 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		return errors.Wrap(err, "could not create runner output")
 	}
 	writer := metadata.NewWriter("kubetest2", junitRunner)
+
+	// var wg sync.WaitGroup
+	// wg.Add(1)
```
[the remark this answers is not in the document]
ShwethaKumbla: Taken care of it 😄

amwat:
/lgtm
/approve

Thanks for working on this!

PR #150, opened 2021-06-30
Return the error directly if the GKE cluster creation fails in the last retryable region/zone

chizhg:


kubetest2-gke/deployer/up.go
```diff
@@ -99,7 +99,10 @@ func (d *Deployer) CreateClusters() error {
 
 		r := retryCount
```
BenTheElder, line 100: why do we have this?

kubetest2-gke/deployer/up.go
```diff
@@ -99,7 +99,10 @@ func (d *Deployer) CreateClusters() error {
...
 		if err := eg.Wait(); err != nil {
-			if d.isRetryableError(err) {
+			// If the error is retryable and it is not the last region/zone that
+			// can be retried, perform cleanups in the background and retry
+			// cluster creation in the next available region/zone.
+			if d.isRetryableError(err) && r != totalTryCount-1 {
 				go func() {
```
BenTheElder, line 106: if r == totalTryCount-1 we're never going to cleanup, right?

go prefers not to indent the common cases

kubetest2-gke/deployer/up.go
BenTheElder: also why is retryCount stored on the deployer?

BenTheElder (commented):
this method has gotten a bit convoluted with a number of nested loops, I will not hard block on this but I recommend breaking it up a bit, e.g. the outer retry loop can call some `tryCreateClusters()` method to cut down on the nested context. 
I'm also not sure why retrycount is being stored redundantly.

kubetest2-gke/deployer/up.go
chizhg: if `r == totalTryCount-1` we will not clean up the clusters asynchronously. Instead, we will clean them up in the `Down` function if `--down` flag is provided. This conforms with the behavior when no retry is needed.

kubetest2-gke/deployer/up.go
chizhg: It's because we also need it for a few other functions, like `d.Up()`, `d.Kubeconfig()` and `d.EnsureFirewallRules()`.

chizhg: That line was added because the Go caveat - https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/. I have moved the code to a separate `tryCreateClusters()` function as per your suggestion, so that issue has gone.

BenTheElder (approved):
/lgtm
/approve

this is more readable, thanks!

kubetest2-gke/deployer/up.go
BenTheElder: that's fair, it seems like the loop should directly increment d.retryCount still though?

I don't see why we have _three_ copies of this value. The capture one is reasonable.

PR #155, opened 2021-07-14
Add windows/amd64 in target platform

anfernee:
The build only builds `linux/amd64` now.

### DIFF
--- a/kubetest2-gke/deployer/build/gke_make.go
+++ b/kubetest2-gke/deployer/build/gke_make.go
@@ -90,7 +90,7 @@ func (gmb *GKEMake) Build() (string, error) {
 
 	// Skip validation for faster builds
 	// TODO: add support for a separate validate mode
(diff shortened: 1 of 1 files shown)


amwat:
/hold

I think this should be set in the config instead of in code

anfernee:
I think for GKE, it's always built this way with support of both platforms. if that's the case, we should probably do the same thing here.

amwat:
From kubetest2 perspective we still want the ability to specify linux only building for faster builds.
Hardcoding the TARGET_PLATFORMS means we lose that. 
Currently, we can pass whatever TARGET_PLATFORMS we want through the config file, we can add windows there.

anfernee:
Which config file are you referring to? I saw some `BuildOptions` fields. Do you think it's common build option or gke specific build option?

amwat:
the `BuildScript` flag takes in a config file with the `gke_make` build options including TARGET_PLATFORMS.

anfernee:
Per @amwat  this change is not needed.

PR #159, opened 2021-07-29
Implemented the update-latest flag for gke_make

richackard:
When the --update-latest flag is set to true, gke_make would upload the build number to gob-prow GCS after the build completes

### DIFF
--- a/kubetest2-gke/deployer/build/gke_make.go
+++ b/kubetest2-gke/deployer/build/gke_make.go
@@ -120,14 +120,18 @@ func (gmb *GKEMake) Stage(version string) error {
 	if gmb.UpdateLatest {
 		r := regexp.MustCompile(gkeMinorVersionRegex)
 		m := r.FindStringSubmatch(version)
+		var fName string
 		if len(m) < 2 {
-			return fmt.Errorf("failed to update the latest version because can't find the minor version of %s", version)
+			klog.Warningf("can't find the minor version of %s, defaulting to latest.txt", version)
+			fName = "latest.txt"
+		} else {
+			minor := m[1]
+			fName = fmt.Sprintf("latest-%s.txt", minor)
 		}
-		minor := m[1]
-		fName := fmt.Sprintf("latest-%s.txt", minor)
-		pushCmd := fmt.Sprintf("echo '%s' | gsutil cp - %s/%s", version, gmb.StageLocation, fName)
-		cmd := exec.Command("bash", "-c", pushCmd)
-		exec.InheritOutput(cmd)
+		pushCmd := fmt.Sprintf("gsutil cp - %s/%s", gmb.StageLocation, fName)
+		cmd := exec.Command(pushCmd)
+		cmd.SetStdin(strings.NewReader(version))
+		exec.SetOutput(cmd, os.Stdout, os.Stderr)
 		if err := cmd.Run(); err != nil {
 			return fmt.Errorf("failed to upload the latest version number: %s", err)
 		}
--- a/kubetest2-gke/deployer/build/gke_make.go
+++ b/kubetest2-gke/deployer/build/gke_make.go
@@ -129,7 +129,7 @@ func (gmb *GKEMake) Stage(version string) error {
 			minor := m[1]
 			fName = fmt.Sprintf("latest-%s.txt", minor)
 		}
-		pushCmd := fmt.Sprintf("gsutil cp - %s/%s", gmb.StageLocation, fName)
+		pushCmd := fmt.Sprintf("gsutil -h 'Content-Type:text/plain' cp - %s/%s", gmb.StageLocation, fName)
 		cmd := exec.RawCommand(pushCmd)
 		cmd.SetStdin(strings.NewReader(version))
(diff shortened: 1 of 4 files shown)


kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -112,6 +116,23 @@ func (gmb *GKEMake) Stage(version string) error {
...
+
+	if gmb.UpdateLatest {
+		r := regexp.MustCompile(gkeMinorVersionRegex)
+		m := r.FindStringSubmatch(version)
+		if len(m) < 2 {
+			return fmt.Errorf("failed to update the latest version because can't find the minor version of %s", version)
+		}
+		minor := m[1]
+		fName := fmt.Sprintf("latest-%s.txt", minor)
+		pushCmd := fmt.Sprintf("echo '%s' | gsutil cp - %s/%s", version, gmb.StageLocation, fName)
```
amwat, line 128: You can set the stdin for the cmd, to avoid echo and bash.

kubetest2-gke/deployer/options/build.go
```diff
@@ -42,6 +42,7 @@ func (bo *BuildOptions) Validate() error {
 					BuildScript:   bo.BuildScript,
 					VersionSuffix: bo.CommonBuildOptions.VersionSuffix,
 					StageLocation: bo.CommonBuildOptions.StageLocation,
+					UpdateLatest:  bo.CommonBuildOptions.UpdateLatest,
```
amwat, line 45: let's also plumb this value through to the KREL stager since it a commonoption ?

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -112,6 +116,23 @@ func (gmb *GKEMake) Stage(version string) error {
 	if err := gmb.runWithActions(os.Stdout, os.Stderr, []gkeBuildAction{stage}, args...); err != nil {
 		return err
 	}
+
+	if gmb.UpdateLatest {
+		r := regexp.MustCompile(gkeMinorVersionRegex)
+		m := r.FindStringSubmatch(version)
+		if len(m) < 2 {
+			return fmt.Errorf("failed to update the latest version because can't find the minor version of %s", version)
```
amwat, line 124: let's make this a warning
and update just `latest.txt` in this case.

richackard: Done

kubetest2-gke/deployer/options/build.go
richackard: Done

kubetest2-gke/deployer/build/gke_make.go
richackard: Done

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -112,6 +116,27 @@ func (gmb *GKEMake) Stage(version string) error {
...
+		var fName string
+		if len(m) < 2 {
+			klog.Warningf("can't find the minor version of %s, defaulting to latest.txt", version)
+			fName = "latest.txt"
+		} else {
+			minor := m[1]
+			fName = fmt.Sprintf("latest-%s.txt", minor)
+		}
+		pushCmd := fmt.Sprintf("gsutil cp - %s/%s", gmb.StageLocation, fName)
+		cmd := exec.Command(pushCmd)
```
amwat, line 132: won't this put everything as the command without any arguments?

you might want exec.RawCommand()

amwat (commented):
Can you try testing this locally end-to-end to see if this works?

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -35,6 +36,8 @@ const (
 	pack         gkeBuildAction = "package"
 	stage        gkeBuildAction = "push-gcs"
 	printVersion gkeBuildAction = "print-version"
+
+	gkeMinorVersionRegex string = "^v(\\d\\.\\d+).*$"
```
MushuEE, line 40: nit: I think we typically keep regex as vars with MustCompile to avoid issue in the regex itself

MushuEE: Also I like `(?<Year>\d{4})-(?<Month>\d{2})-(?<Day>\d{2})` using the named capture groups. Really helps understand what is what. and you don't need to just index the match. you can say match["minor_version"]

MushuEE (commented):
Some nits but will leave to Amit to lgtm

amwat (commented):
/lgtm
/approve

/hold
feel free to unhold after testing locally and if you plan to make the named capture group changes.

kubetest2-gke/deployer/build/gke_make.go
richackard: > nit: I think we typically keep regex as vars with MustCompile to avoid issue in the regex itself
Done.
> named capture group
I think the named group is not needed in this case since the variable name has implied what it is.

kubetest2-gke/deployer/build/gke_make.go
richackard: Done

PR #160, opened 2021-08-02
Allow gce deployer to use make instead of Bazel

jiahuif:
We are now switching kubernetes/cloud-provider-gcp to use GNU make instead of Bazel as the build system, analog to the main k/k repository. This PR allows the gce deployer to look for hint within the repository to choose from make and Bazel

### DIFF
--- a/kubetest2-gce/deployer/build.go
+++ b/kubetest2-gce/deployer/build.go
@@ -19,6 +19,7 @@ package deployer
 import (
 	"fmt"
 	"os"
+	"path"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


spiffxp (approved):
/lgtm
/hold
/assign @amwat 
for the /approve and /hold cancel

amwat (commented):
for the k/k provider we have specific builders which we choose via a command line flag.
Not sure if we want to do something like that here, but current prioritization of make over bazel lgtm.

/approve
/hold cancel

PR #161, opened 2021-08-02
Change the gke minor regex to use single backslash to escape since it's in a raw string.

richackard:


kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -44,7 +44,7 @@ const (
 )
 
 var (
-	gkeMinorVersionRegex = regexp.MustCompile(`^v(\\d\\.\\d+).*$`)
+	gkeMinorVersionRegex = regexp.MustCompile(`^v(\d\.\d+).*$`)
```
amwat, line 47: let's reuse the regex at
https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gke/deployer/version.go#L43

we already have a few unit tests for that

richackard: The capture group in that regex is a bit tricky to get information from

amwat: ACK, let's keep it simple here then.

amwat (commented):
/lgtm
/approve

PR #162, opened 2021-08-07
Add version support for all binaries

amwat:
also add support for writing `metadata.json` similar to what kubetest does in existing prow jobs.

fixes: https://github.com/kubernetes-sigs/kubetest2/issues/108

For deployers:
the shim GitTag won't be present for out-of-tree providers instead inject an env `KUBETEST2_VERSION` from the shim to the deployers.

For testers:
Went the route of overwriting metadata in testers.
alternative was to enforce all testers support a `--version` flag and detect it similar to `--help`, but is unnecessarily restrictive.

/cc @BenTheElder @spiffxp

### DIFF
--- a/kubetest2-gke/deployer/common.go
+++ b/kubetest2-gke/deployer/common.go
@@ -44,6 +44,10 @@ func (d *Deployer) Init() error {
 
 // Initialize should only be called by init(), behind a sync.Once
 func (d *Deployer) Initialize() error {
+	if d.ClusterVersion == "" && d.LegacyClusterVersion != "" {
+		klog.Warningf("--version is deprecated please use --cluster-version")
+		d.ClusterVersion = d.LegacyClusterVersion
+	}
 	if d.kubetest2CommonOptions.ShouldUp() {
 		d.totalTryCount = math.Max(len(d.Regions), len(d.Zones))
 
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -38,6 +38,8 @@ import (
 // Name is the name of the deployer
 const Name = "gke"
 
+var GitTag string
+
 const (
 	e2eAllow            = "tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767"
 	defaultImage        = "cos"
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -146,6 +146,10 @@ func (d *Deployer) Provider() string {
 	return Name
 }
 
+func (d *Deployer) Version() string {
+	return GitTag
+}
+
 // New implements deployer.New for gke
 func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 	// create a deployer object and set fields that are not flag controlled
@@ -171,7 +175,7 @@ func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 			WindowsMachineType: defaultWindowsNodePool.MachineType,
 			WindowsImageType:   defaultWindowsImage,
 			// Leave Version as empty to use the default cluster version.
-			Version:          "",
+			ClusterVersion:   "",
 			GCPSSHKeyIgnored: true,
 
 			BoskosLocation:                 defaultBoskosLocation,
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -174,7 +174,7 @@ func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 			WindowsNumNodes:    defaultWindowsNodePool.Nodes,
 			WindowsMachineType: defaultWindowsNodePool.MachineType,
 			WindowsImageType:   defaultWindowsImage,
-			// Leave Version as empty to use the default cluster version.
+			// Leave ClusterVersion as empty to use the default cluster version.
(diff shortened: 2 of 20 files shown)


amwat:
I would recommend reviewing by commit

kubetest2-gke/deployer/common.go
```diff
@@ -44,6 +44,10 @@ func (d *Deployer) Init() error {
 
 // Initialize should only be called by init(), behind a sync.Once
 func (d *Deployer) Initialize() error {
+	if d.ClusterVersion == "" && d.LegacyClusterVersion != "" {
+		klog.Warningf("--version is deprecated please use --cluster-version")
```
amwat, line 48: fyi @chizhg

pkg/app/app.go
```diff
@@ -159,3 +163,39 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
+	)
+	if err != nil {
+		return err
+	}
+
+	meta, err2 := metadata.NewCustomJSON(nil)
+	if err2 != nil {
+		return err2
+	}
+	if err := meta.Add("kubetest-version", os.Getenv("KUBETEST2_VERSION")); err != nil {
```
amwat, line 180: this currently tries to match with
https://github.com/kubernetes/test-infra/issues/20886#issuecomment-894586227

BenTheElder (approved):
/hold

kubetest2-gke/deployer/deployer.go
```diff
@@ -169,7 +175,7 @@ func New(opts types.Options) (types.Deployer, *pflag.FlagSet) {
 			WindowsMachineType: defaultWindowsNodePool.MachineType,
 			WindowsImageType:   defaultWindowsImage,
 			// Leave Version as empty to use the default cluster version.
```
chizhg, line 177: Comment also needs to be updated to `ClusterVersion`

kubetest2-gke/deployer/common.go
chizhg: ACK

kubetest2-gke/deployer/deployer.go
spiffxp: +1

kubetest2-gke/deployer/options/up.go
```diff
@@ -29,7 +29,8 @@ type UpOptions struct {
 	NumNodes                int    `flag:"~num-nodes" desc:"For use with gcloud commands to specify the number of nodes for the cluster."`
 	ImageType               string `flag:"~image-type" desc:"The image type to use for the cluster."`
 	ReleaseChannel          string `desc:"Use a GKE release channel, could be one of empty, rapid, regular and stable - https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels"`
-	Version                 string `desc:"Use a specific GKE version e.g. 1.16.13.gke-400, 'latest' or ''. If --build is specified it will default to building kubernetes from source."`
+	ClusterVersion          string `desc:"Use a specific GKE version e.g. 1.16.13.gke-400, 'latest' or ''. If --build is specified it will default to building kubernetes from source."`
+	LegacyClusterVersion    string `flag:"~version,deprecated" desc:"Use --cluster-version instead"`
```
spiffxp, line 33: When will this get removed?

pkg/app/app.go
```diff
@@ -159,3 +163,39 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 	}
 	return nil
 }
+
+func writeVersionToMetadataJSON(opts types.Options, d types.Deployer) error {
+	// setup the json metadata writer
+	metadataJSON, err := os.Create(
+		filepath.Join(opts.RunDir(), "metadata.json"),
+	)
```
spiffxp, line 171: Any reason this can't match the behavior of `pkg.testers.metadata.WriteVersionToMetadata` and preserve an existing metadata if it happens to exist?

pkg/types/types.go
```diff
@@ -113,6 +113,14 @@ type DeployerWithPostTester interface {
...
 }
 
+// DeployerWithVersion allows the deployer to specify it's version
+type DeployerWithVersion interface {
+	Deployer
+
+	// Version determines the version of the deployer binary
+	Version() string
+}
+
```
spiffxp, line 123: What would be involved in making `Version()` part of `Deployer`?  Seeing three different `DeployerWithFoo` interfaces makes me wonder if it keeps things simpler to fold into core

spiffxp (changes requested):
Overall LGTM, mostly questions

The change I'm requesting is the one @chizhg pointed out

kubetest2-gke/deployer/deployer.go
amwat: done

kubetest2-gke/deployer/options/up.go
amwat: most likely in the release afterwards. (v0.2.0)

pkg/app/app.go
amwat: It can, but atleast at this point kubetest2 itself doesn't expect any `metadata.json` to be written out to the run specific output directory by itself.

pkg/types/types.go
amwat: I think that's something we can add in v0.2.0
Currently, all out of tree deployers won't have `Version()` and will break the core interface.

spiffxp (approved):
/approve
/lgtm

PR #170, opened 2021-10-29
adding `image-config-dir` to node tester

Namanl2001:
Signed-off-by: Namanl2001 <namanlakhwani@gmail.com>

adding `image-config-dir` to node tester

cc: @amwat @dims


amwat (commented):
/ok-to-test
/lgtm
/approve

Looks good for now, in the future we should probably separate out flags that are associated only with run_remote into a separate struct.

PR #171, opened 2021-11-04
gke: Avoid hardcoding COS as default image type

bobbypage:
The default image type in GKE as of 1.19 is COS_CONTAINERD. kubetest2
gke deployer should avoid hardcoding the image and instead rely on the
default unless image is specified.

### DIFF
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -42,7 +42,6 @@ var GitTag string
 
 const (
 	defaultFirewallRuleAllow = "tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767"
-	defaultImage             = "cos"
 	defaultWindowsImage      = WindowsImageTypeLTSC
 )
 
@@ -197,7 +196,6 @@ func NewDeployer(opts types.Options) *Deployer {
 			NumClusters: 1,
 			NumNodes:    defaultNodePool.Nodes,
 			MachineType: defaultNodePool.MachineType,
-			ImageType:   defaultImage,
 			// Leave ClusterVersion as empty to use the default cluster version.
 			ClusterVersion:    "",
 			FirewallRuleAllow: defaultFirewallRuleAllow,
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -42,7 +42,6 @@ var GitTag string
 
 const (
 	defaultFirewallRuleAllow = "tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767"
-	defaultWindowsImage      = WindowsImageTypeLTSC
 )
 
 const (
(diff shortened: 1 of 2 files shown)


kubetest2-gke/deployer/deployer.go
```diff
@@ -42,7 +42,6 @@ var GitTag string
 
 const (
 	defaultFirewallRuleAllow = "tcp:22,tcp:80,tcp:8080,tcp:30000-32767,udp:30000-32767"
-	defaultImage             = "cos"
 	defaultWindowsImage      = WindowsImageTypeLTSC
```
chizhg, line 45: Shall we also remove the `defaultWindowsImage`, for consistency reason?

chizhg (commented):
/ok-to-test

kubetest2-gke/deployer/deployer.go
bobbypage: sounds good, followed up with second commit to remove windows image defaulting.

bobbypage:
Failure in CI is unrelated panic...

```
+ kubetest2 gke -v 2 --num-clusters 1 --num-nodes 1 --zone us-central1-c,us-west1-a,us-east1-b --network ci-tests-network --up --down --test=exec -- /home/prow/go/src/github.com/kubernetes-sigs/kubetest2/kubetest2-gke/ci-tests/test.sh
I1104 00:24:50.106030    4445 app.go:59] RunDir for this run: "/logs/artifacts/8957be9c-3d05-11ec-bb62-5a9d54b04110"
I1104 00:24:50.126348    4445 app.go:94] ID for this run: "8957be9c-3d05-11ec-bb62-5a9d54b04110"
I1104 00:24:50.126478    4445 common.go:69] No GCP projects provided, acquiring from Boskos [1] project/s
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
sigs.k8s.io/kubetest2/kubetest2-gke/deployer.(*Deployer).Down(0xc000144100, 0xc0079c00a0, 0x26b2ec2f5fc65)
	sigs.k8s.io/kubetest2/kubetest2-gke/deployer/down.go:35 +0x2da
sigs.k8s.io/kubetest2/pkg/metadata.(*Writer).WrapStep(0xc00013c090, 0xe0385a, 0x4, 0xc0003497a8, 0xc000000180, 0xe4b2f0)
	sigs.k8s.io/kubetest2/pkg/metadata/writer.go:51 +0x70
```

qiutongs (approved):

SergeyKanzhelev (approved):
This is great. I'd say we may also want to apply the same for the machine type. Current default in kubetest is not matching the GKE default

SergeyKanzhelev:
Also filed https://github.com/kubernetes-sigs/gcp-filestore-csi-driver/issues/186

bobbypage:
> This is great. I'd say we may also want to apply the same for the machine type. Current default in kubetest is not matching the GKE default

Thanks @SergeyKanzhelev. Agree, I think the less we default on the client and rely on the server side defaults the better. @amwat what do you think about removing default for machine type as well?

amwat:
/approve

I agree, let's rely on server defaults as much as possible.

PR #172, opened 2021-11-04
Some updates to kubetest2 gke deployer

chizhg:
1. Change more fields to be public to use it as a library
2. For multi-project multi-cluster topology, allow to not specify the project index in the cluster names, in which case the cluster index will be the same as the project index
3. It's not useful to print the number of firewall rules deleted, so dropping the calculation and return logic

### DIFF
--- a/kubetest2-gke/deployer/deployer.go
+++ b/kubetest2-gke/deployer/deployer.go
@@ -89,10 +89,10 @@ type ig struct {
 	uniq string
 }
 
-type cluster struct {
-	// index is the index of the cluster in the list provided via the --cluster-name flag
-	index int
-	name  string
+type Cluster struct {
+	// Index is the index of the cluster in the list provided via the --cluster-name flag
+	Index int
+	Name  string
 }
 
 type Deployer struct {
@@ -108,7 +108,7 @@ type Deployer struct {
 	// doInit helps to make sure the initialization is performed only once
 	doInit sync.Once
 	// only used for multi-project multi-cluster profile to save the project-clusters mapping
-	projectClustersLayout map[string][]cluster
+	projectClustersLayout map[string][]Cluster
 	// project -> cluster -> instance groups
 	instanceGroups map[string]map[string][]*ig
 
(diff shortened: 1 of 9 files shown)


chizhg:
/assign @BenTheElder @MushuEE 
Could you help review this PR? Thanks!

chizhg:
It's already used as a library - we are using it to implement an internal tool for Anthos, Amit reviewed all the previous PRs and this is changing more fields to be public.

We are also printing the errors, so printing the number of deleted firewall rules feels a bit redundant, but I can add it back if you want it.

And sorry for bothering..., I'm now mostly wrapping up stuff and cleaning up some tech debt, so this should be my last change to this repo...

BenTheElder:
> It's already used as a library - we are using it to implement an internal tool for Anthos, Amit reviewed all the previous PRs and this is changing more fields to be public.

I see that a few functions were made public previously.

I probably would have disagreed with this change -- deployer already has a standard public API, but also those changes are not the same as exposing types like the internal cluster index struct.

> We are also printing the errors, so printing the number of deleted firewall rules feels a bit redundant, but I can add it back if you think it's more desirable.

I think it's useful to know the number. I don't think it's clearly _not_ useful to know it.

> [Bug fix] Boskos release should set the resource state to dirty

This seems like it should really be a separate PR.

> For multi-project multi-cluster topology, allow to not specify the project index in the cluster names, in which case the cluster index will be the same as the project index

Have not had time to review this aspect, sorry, currently working on KETTLE outage .. this repo is low-prio bugfix focused right now until we can re-staff.

kubetest2-gke/deployer/deployer.go
```diff
@@ -89,10 +89,10 @@ type ig struct {
 	uniq string
 }
 
-type cluster struct {
-	// index is the index of the cluster in the list provided via the --cluster-name flag
-	index int
-	name  string
+type Cluster struct {
+	// Index is the index of the cluster in the list provided via the --cluster-name flag
+	Index int
```
BenTheElder, line 94: this is the sort of internal detail that is not reasonable to export

BenTheElder (commented):
These types should not be exposed, the deployer interface is the public API.

PR #173, opened 2021-11-04
Use gce-project to run e2e tests for kubetest2 gke deployer

chizhg:
Since `gke-project` pool was removed from https://github.com/kubernetes/test-infra/pull/23880

Also fix the panic when it fails to acquire projects from Boskos - https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kubetest2/171/pull-kubetest2-gke-up-down-singlecluster/1456054639957905408


spiffxp (approved):
/approve
/lgtm

PR #174, opened 2021-11-04
gke: Remove hardcoded default machine types

bobbypage:
Remove hardcoded machine types and instead rely on the server side
defaults. Only pass machineType flag if machine type is explicitly
provided.


bobbypage:
Followup on @SergeyKanzhelev comment here - https://github.com/kubernetes-sigs/kubetest2/pull/171#pullrequestreview-797263709

SergeyKanzhelev (approved):
thank you for the follow up!

amwat:
/approve

Thank you!

PR #175, opened 2021-11-15
Configuring --gcp-project-type and --runtime-config

Namanl2001:
Configuring `--gcp-project-type` and `--runtime-config`

cc: @amwat @dims

### DIFF
--- a/pkg/testers/node/node.go
+++ b/pkg/testers/node/node.go
@@ -41,10 +41,9 @@ import (
 var GitTag string
 
 const (
-	target                 = "test-e2e-node"
-	gceProjectResourceType = "gce-project"
-	ciPrivateKeyEnv        = "GCE_SSH_PRIVATE_KEY_FILE"
-	ciPublicKeyEnv         = "GCE_SSH_PUBLIC_KEY_FILE"
+	target          = "test-e2e-node"
+	ciPrivateKeyEnv = "GCE_SSH_PRIVATE_KEY_FILE"
+	ciPublicKeyEnv  = "GCE_SSH_PUBLIC_KEY_FILE"
 )
 
 type Tester struct {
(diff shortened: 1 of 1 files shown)


Namanl2001:
Corresponding PR in K/K https://github.com/kubernetes/kubernetes/pull/106431

pkg/testers/node/node.go
```diff
@@ -61,6 +60,8 @@ type Tester struct {
 	ImageConfigFile                string `desc:"Path to a file containing image configuration."`
 	ImageConfigDir                 string `desc:"Path to image config files."`
 	Parallelism                    int    `desc:"The number of nodes to run in parallel."`
+	GcpProjectType                 string `desc:"Explicitly indicate which project type to select from boskos."`
```
amwat, line 63: nit: use `GCPProjectType` instead of `GcpProjectType`

pkg/testers/node/node.go
```diff
@@ -242,6 +244,7 @@ func (t *Tester) constructArgs() []string {
 		"IMAGE_CONFIG_DIR=" + t.ImageConfigDir,
 		"SSH_USER=" + t.sshUser,
 		"SSH_KEY=" + t.privateKey,
+		"RUNTIME_CONFIG=" + t.RuntimeConfig,
```
amwat, line 247: does it work correctly if it empty?

amwat (commented):
/lgtm
/approve
/hold

pkg/testers/node/node.go
Namanl2001: Thanks for catching it out. This needs to be checked explicitly as it's been done for kubetest https://github.com/kubernetes/test-infra/blob/8c944aa6aa17c1efd224629b250e9bd52eeb7132/kubetest/e2e.go#L625-L627

pkg/testers/node/node.go
Namanl2001: 👍 output would be same for both, but `GCPProjectType` seems nicer.

PR #177, opened 2021-12-13
kubetest2 gke --down releases Boskos projects if they are acquired from Boskos

chizhg:

PR #180, opened 2022-01-28
[Bug fix] Boskos release should set the resource state to dirty

chizhg:
To use Boskos as the GCP project rental, when releasing the project, it should reset its state to `dirty` instead of `free`, this'll allow GCP Janitor to clean up all resources in the project and reset it to `free`, so that the project can be returned to the project rental in a clean state.

/cc @MushuEE

### DIFF
--- a/pkg/boskos/boskos.go
+++ b/pkg/boskos/boskos.go
@@ -95,7 +95,7 @@ func startBoskosHeartbeat(boskosClient *client.Client, resource *common.Resource
 // Release releases a resource.
 func Release(client *client.Client, resourceNames []string, heartbeatClose chan struct{}) error {
 	for _, name := range resourceNames {
-		if err := client.Release(name, "free"); err != nil {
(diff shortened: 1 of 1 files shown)


chizhg:
/cc @BenTheElder 
Hi Ben, Aaron, could you help review this small PR? It's only a one-line change. Thanks!

BenTheElder:
Aaron is on long-term leave ATM.

I'm concerned that we've been running this and apparently not noticing 🤨 
/lgtm
/hold

chizhg:
> Aaron is on long-term leave ATM.
> 
> I'm concerned that we've been running this and apparently not noticing 🤨 /lgtm /hold

Previously `--down` did not release the Boskos projects and I made the change in https://github.com/kubernetes-sigs/kubetest2/pull/177

chizhg:
It's actually missing /approve

PR #181, opened 2022-01-30
Clarify error message when pre-built binary not found

justinsb:
This makes usage of pre-built binaries more intuitive.


BenTheElder (approved):
/lgtm
/approve

PR #182, opened 2022-02-07
Allow rundir to be set with SetRunDir

nckturner:
Kops wraps the ginkgo tester with its own tester, but doesn't has its own `Execute()` method.  This will allow the kops tester to use `KUBETEST2_RUN_DIR` as the `runDir` without much refactoring. 

See https://github.com/kubernetes/kops/pull/13217


nckturner:
Is this reasonable?

BenTheElder:
sorry, Aaron is on leave and Amit moved on so it's just me in this repo right now, this seems reasonable.

BenTheElder (approved):

Issue #98, opened 2021-03-01
Decouple RunDir from Artifacts


amwat:
A tracking issue for: https://github.com/kubernetes-sigs/kubetest2/pull/91#pullrequestreview-591825637

RunDir: should be used for storing all files in general specific to a single run of kubetest2
Artifacts: should only be used intentionally for storing files which (in CI) get automatically uploaded to GCS. 

also helps with: https://github.com/kubernetes-sigs/kubetest2/issues/97

/cc @justinsb @mkumatag 

The logical separation was added in: https://github.com/kubernetes-sigs/kubetest2/pull/91
But currently RunDir (by default) is a subdirectory under artifacts, so there's no actual separation yet.

Rajalakshmi-Girish:
@amwat I have started working on this. Below is the approach that I thought of. Can you please confirm if this seems good to proceed?

- Have RunDir out of Artifacts and decoupled as this issue aims. (Not a subdirectory. Default RunDir shall be `_rundir` at the same level as `_artifacts`)

- All the files related to a specific run(Binaries, metadata.json, junit_runner.xml and junit_*.xml files) would be initially be stored in RunDir.

- Later, the junit files would be copied to Artifacts directory once the run completes. (Copy of junit files from RunDir to Artifacts would happen unconditionally)

- A new flag `upload-test-binaries`  would be introduced that would be `False` by default. If this flag is passed `True` in the `kubetest2` command, the binaries (gingko, e2e.test, kubectl) would be copied to Artifacts from Rundir.
Else the binaries would be left only in Rundir and not be uploaded to storage.

Rajalakshmi-Girish:
> @amwat I have started working on this. Below is the approach that I thought of. Can you please confirm if this seems good to proceed?
> 
>     * Have RunDir out of Artifacts and decoupled as this issue aims. (Not a subdirectory. Default RunDir shall be `_rundir` at the same level as `_artifacts`)
> 
>     * All the files related to a specific run(Binaries, metadata.json, junit_runner.xml and junit_*.xml files) would be initially be stored in RunDir.
> 
>     * Later, the junit files would be copied to Artifacts directory once the run completes. (Copy of junit files from RunDir to Artifacts would happen unconditionally)
> 
>     * A new flag `upload-test-binaries`  would be introduced that would be `False` by default. If this flag is passed `True` in the `kubetest2` command, the binaries (gingko, e2e.test, kubectl) would be copied to Artifacts from Rundir.
[...]

PR #183, opened 2022-03-03
Decouple RunDir from Artifacts

Rajalakshmi-Girish:
Fixes https://github.com/kubernetes-sigs/kubetest2/issues/98
This change separates RunDir from Artifacts and introduces a flag `cp-rundir-to-artifacts` for intentionally copying binaries/ metadata to artifacts.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/klog"
 
+	"sigs.k8s.io/kubetest2/pkg/artifacts"
 	"sigs.k8s.io/kubetest2/pkg/exec"
 	"sigs.k8s.io/kubetest2/pkg/metadata"
 	"sigs.k8s.io/kubetest2/pkg/types"
@@ -57,7 +58,10 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		Throughout this, collecting metadata and writing it out on exit
 	*/
 	// TODO(bentheelder): signal handling & timeout
-
+	if !opts.RundirInArtifacts() {
+		klog.Infof("The files in RunDir shall not be part of Artifacts")
+		klog.Infof("pass rundir-in-artifacts flag True for RunDir to be part of Artifacts")
+	}
 	klog.Infof("RunDir for this run: %q", opts.RunDir())
 
 	// ensure the run dir
@@ -69,9 +73,14 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		return err
 	}
 
+	// ensure the artifacts dir
+	if err := os.MkdirAll(artifacts.BaseDir(), os.ModePerm); err != nil {
+		return err
+	}
+
 	// setup junit writer
 	junitRunner, err := os.Create(
-		filepath.Join(opts.RunDir(), "junit_runner.xml"),
+		filepath.Join(artifacts.BaseDir(), "junit_runner.xml"),
 	)
 	if err != nil {
 		return errors.Wrap(err, "could not create runner output")
@@ -158,7 +167,7 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		// also add run_dir to $PATH for locally built binaries
 		updatedPath := opts.RunDir() + string(filepath.ListSeparator) + os.Getenv("PATH")
 		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "PATH", updatedPath))
-		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "ARTIFACTS", opts.RunDir()))
+		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "ARTIFACTS", artifacts.BaseDir()))
 		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "KUBETEST2_RUN_DIR", opts.RunDir()))
 		envsForTester = append(envsForTester, fmt.Sprintf("%s=%s", "KUBETEST2_RUN_ID", opts.RunID()))
 		// If the deployer provides a kubeconfig pass it to the tester
--- a/pkg/app/cmd.go
+++ b/pkg/app/cmd.go
@@ -191,6 +191,7 @@ type options struct {
 	test                string
 	skipTestJUnitReport bool
 	runid               string
+	rundirInArtifacts   bool
 }
 
 // bindFlags registers all first class kubetest2 flags
@@ -202,7 +203,6 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&o.test, "test", "", "test type to run, if unset no tests will run")
 	flags.BoolVar(&o.skipTestJUnitReport, "skip-test-junit-report", false, "skip reporting the test step as a JUnit test case, "+
 		"should be set to true when solely relying on the tester binary to generate it's own junit.")
-
 	var defaultRunID string
 	// reuse uid for CI use cases
 	if uid, exists := os.LookupEnv("PROW_JOB_ID"); exists && uid != "" {
@@ -211,6 +211,7 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 		defaultRunID = uuid.New().String()
 	}
 	flags.StringVar(&o.runid, "run-id", defaultRunID, "unique identifier for a kubetest2 run")
+	flags.BoolVar(&o.rundirInArtifacts, "rundir-in-artifacts", false, `if true, the test binaries and run specific metadata will be in the ARTIFACTS`)
 }
 
 // assert that options implements deployer options
@@ -245,7 +246,26 @@ func (o *options) RunID() string {
 }
 
 func (o *options) RunDir() string {
-	return filepath.Join(artifacts.BaseDir(), o.RunID())
+	if o.RundirInArtifacts() {
+		//making rundir under ARTIFACTS
+		return filepath.Join(artifacts.BaseDir(), subRunDir(), o.RunID())
+	}
+	return filepath.Join(artifacts.RunDir(), o.RunID())
+}
+
+func subRunDir() string {
+	//Function to make sure only relative path is appended to artifacts BaseDir
+	if artifacts.RunDirFlag == "" {
+		if path, set := os.LookupEnv("KUBETEST2_RUN_DIR"); set {
+			return path
+		}
+		return "_rundir"
+	}
+	return artifacts.RunDirFlag
+}
+
+func (o *options) RundirInArtifacts() bool {
+	return o.rundirInArtifacts
 }
 
 // metadata used for CLI usage string
--- a/pkg/artifacts/paths.go
+++ b/pkg/artifacts/paths.go
@@ -18,14 +18,14 @@ package artifacts
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
-
 	"github.com/spf13/pflag"
 	"k8s.io/klog"
+	"os"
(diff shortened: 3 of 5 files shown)


Rajalakshmi-Girish:
@mkumatag ^^

pkg/artifacts/paths.go
```diff
@@ -41,6 +41,20 @@ func BaseDir() string {
 	return d
 }
 
+// RunDir return the path to the directory used for storing all files in general
```
MushuEE, line 44:
```suggestion
// RunDir returns the path to the directory used for storing all files in general
```

pkg/artifacts/paths.go
```diff
@@ -41,6 +41,20 @@ func BaseDir() string {
...
+// specific to a single run of kubetest2
+func RunDir() string {
+	d := runDir
+	if d == "" {
```
MushuEE, line 48: I see this was copied from above, but since `d` is being defaulted to s string type, is it not always ""?

pkg/artifacts/paths.go
```diff
@@ -60,20 +74,44 @@ func defaultArtifactsDir() (string, error) {
...
+	flags.StringVar(&runDir, "rundir", defaultRundir, `directory to put run related test binaries like e2e.test, ginkgo, kubectl for each kubetest2 run, defaulting to "${KUBETEST2_RUN_DIR:-./_rundir}". If using the ginkgo tester, this must be an absolute path.`)
 	return nil
 }
 
 // MustBindFlags calls BindFlags and panics on an error
 func MustBindFlags(flags *pflag.FlagSet) {
 	err := BindFlags(flags)
 	if err != nil {
-		klog.Fatalf("failed to get default artifacts directory: %s", err)
+		klog.Fatalf("failed to get default artifacts/rundir directory: %s", err)
```
MushuEE, line 115:
```suggestion
		klog.Fatalf("failed to get default artifacts || rundir directory: %s", err)
```
`artifacts/rundir` sounds like it itself could be a dir haha

MushuEE (commented):
/LGTM

@BenTheElder would love if you could also take a look.

pkg/app/app.go
```diff
@@ -69,9 +70,14 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
 	}
 
+	// ensure the artifacts dir
+	if err := os.MkdirAll(artifacts.BaseDir(), os.ModePerm); err != nil {
+		return err
+	}
+
 	// setup junit writer
 	junitRunner, err := os.Create(
-		filepath.Join(opts.RunDir(), "junit_runner.xml"),
```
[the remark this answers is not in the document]
Rajalakshmi-Girish: Currently, `RunDir` is a sub directory to ARTIFACTS directory. Hence the above location of `junit_runner.xml` in RunDir would definitely upload the file to ARTIFACTS.

But as this change is to decouple RunDir from Artifacts, and file `junit_runner.xml` is an artifact file rather than one specific to each run of kubetest2, hence I have made the change of location from `opts.RunDir()` to `artifacts.BaseDir()`

pkg/artifacts/paths.go
Rajalakshmi-Girish: d fetches its value from `runDir` variable, which indeed is assigning the value passed to flag `--rundir`

Ref: https://github.com/kubernetes-sigs/kubetest2/pull/183/files#diff-76b7e3fd92da504af77a74d8e8e2f888630ee1005503eb1feb04eb30e9b5c172R107

Thus the default value of `d` is not always "". It is "" only when no value is passed to it while running `kubetest`.

Rajalakshmi-Girish (commented):
@MushuEE

pkg/artifacts/paths.go
Rajalakshmi-Girish: Done :)

pkg/artifacts/paths.go
Rajalakshmi-Girish: Done :)

Rajalakshmi-Girish:
@spiffxp Can you please take a look when you have some time?

BenTheElder:
spiffxp is not available ATM and won't be for some time.

Rajalakshmi-Girish:
> spiffxp is not available ATM and won't be for some time.

Thanks for letting us know.
Do you have some time to review this, please? :)

pkg/app/app.go
```diff
@@ -167,6 +173,13 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 			testErr = test.Run()
 		}
 
+		if opts.UploadRundir() {
+			cmd := exec.Command("cp", "-r", artifacts.RunDir(), artifacts.BaseDir())
```
BenTheElder, line 177: // TODO: copy in golang?

Worth noting that actually mimicing cp -r correctly takes some effort.
KIND has a OK implementation at https://github.com/kubernetes-sigs/kind/blob/de5b3e615a4f8b03d2f525e9faf397b12f8a2339/pkg/fs/fs.go#L57

pkg/app/cmd.go
```diff
@@ -211,6 +211,7 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 		defaultRunID = uuid.New().String()
 	}
 	flags.StringVar(&o.runid, "run-id", defaultRunID, "unique identifier for a kubetest2 run")
+	flags.BoolVar(&o.uploadRundir, "upload-rundir", false, `To decide whether the test binaries and run specific metadata have to be uploaded to the ARTIFACTS `)
```
BenTheElder, line 214: I think we should rename this, `kubetest2` does not "upload" anything itself.

Also: perhaps we should have an option to simply have rundir be under artifacts again? I'm not sure what that should look like without thinking about it some more ...

pkg/artifacts/paths.go
```diff
@@ -18,14 +18,14 @@ package artifacts
...
-	"path/filepath"
-
 	"github.com/spf13/pflag"
 	"k8s.io/klog"
+	"os"
+	"path/filepath"
 )
 
 var baseDir string
+var runDir string
```
BenTheElder, line 28: one of these days we should really ditch the globals, but this was already the pattern

pkg/artifacts/paths.go
```diff
@@ -18,14 +18,14 @@ package artifacts
 
 import (
 	"fmt"
-	"os"
-	"path/filepath"
-
 	"github.com/spf13/pflag"
 	"k8s.io/klog"
+	"os"
```
BenTheElder, line 23: nit: stdlib imports should be grouped together at the top, these should not have moved down with third party imports.

it's not _super_ important, but it's the pattern in kubernetes projects, it helps make it clear what the external dependencies are and keeping it this way minimizes the diff

pkg/app/cmd.go
Rajalakshmi-Girish: @BenTheElder How about renaming it to `cpRundirToArtifacts`? (Please suggest any better name for the flag)

Also: Yes I did the same. When the flag `uploadRundir` was set to true, I was copying artifacts.RunDir() to artifacts.BaseDir()
https://github.com/kubernetes-sigs/kubetest2/blob/49251e4560b1c5c51eccb0458d0f8a83ea880763/pkg/app/app.go#L177

pkg/artifacts/paths.go
Rajalakshmi-Girish: @BenTheElder Can we leave this as globals for now?

pkg/app/app.go
mkumatag: That's not package meant be imported outside that project, it's better to copy here the logic.

Rajalakshmi-Girish: There is already an fs package in `kubetest2`. Copied missing functions for directory copy from `kind` repo.

Rajalakshmi-Girish:
@BenTheElder I have incorporated your comments from the previous review.

**These changes were verified for the following scenarios:**

When environment variable `KUBETEST2_RUN_DIR` is set 
```
1. `cp-rundir-to-artifacts` flag is set to `True`
2. `cp-rundir-to-artifacts` flag is not set.
```
`KUBETEST2_RUN_DIR` is not set 
```
3. value passed in flag `--rundir` with `cp-rundir-to-artifacts` flag is set to `True`
4. value passed in flag `--rundir` with `cp-rundir-to-artifacts` flag is not set.
```
When `KUBETEST2_RUN_DIR` is not set and `--rundir` is not passed.
```
5. cp-rundir-to-artifacts` flag is set to `True`
6. `cp-rundir-to-artifacts` flag is not set.
```

@mkumatag ^^

pkg/app/cmd.go
```diff
@@ -211,6 +211,7 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 		defaultRunID = uuid.New().String()
 	}
 	flags.StringVar(&o.runid, "run-id", defaultRunID, "unique identifier for a kubetest2 run")
+	flags.BoolVar(&o.cpRundirToArtifacts, "cp-rundir-to-artifacts", false, `To decide whether the test binaries and run specific metadata have to be copied to the ARTIFACTS `)
```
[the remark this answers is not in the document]
Rajalakshmi-Girish: @BenTheElder I reframed by removing the additional copying.
`rundir` would be under ARTIFACTS Only when `--cp-rundir-to-artifacts` is true.

Did this by joining the RunDir to artifacts BaseDir when the copy flag is true.
Had to create a new function `subRunDir` to avoid rundir with absolute path when it is created under ARTIFACTS.

Please take a look.

pkg/app/cmd.go
```diff
@@ -211,6 +211,7 @@ func (o *options) bindFlags(flags *pflag.FlagSet) {
 		defaultRunID = uuid.New().String()
 	}
 	flags.StringVar(&o.runid, "run-id", defaultRunID, "unique identifier for a kubetest2 run")
+	flags.BoolVar(&o.cpRundirToArtifacts, "cp-rundir-to-artifacts", false, `if true, copy the test binaries and run specific metadata to the ARTIFACTS`)
```
BenTheElder, line 214: this should be renamed? since there's no `cp` happening now

(and the related fields)

otherwise, seems fine now 🙏

mkumatag: @Rajalakshmi-Girish is on medical leave for next couple of week, will relook into the PR once she is back

BenTheElder: Take care!!

pkg/app/app.go
```diff
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"sigs.k8s.io/kubetest2/pkg/artifacts"
```
BenTheElder, line 24: non-blocking nit: typically in Kubernetes we group imports like this:

stdlib

third party packages

packages from the same project

BenTheElder (commented):
quick skim general LGTM,

do you mind rebasing to fixup/squash / avoid the merge commit? just to keep the history clean.

something like `git rebase -i HEAD~2`, replace pick => f (for fixup) for the second commit

/approve

BenTheElder:
thanks for working through figuring out the best approach on this and keeping after the PR, even with all the delayed reviews etc 🙏 

still looking to get more folks involved in reviewing and approving here, I may have to refocus soon 😅

Rajalakshmi-Girish:
> do you mind rebasing to fixup/squash / avoid the merge commit? just to keep the history clean.

I am facing it difficult to rebase and squash those commits :(
Not sure why, but `git rebase -i HEAD~2` is not displaying the commit that I want to squash.
Will check this tomorrow.

BenTheElder (approved):
/lgtm
/approve

sorry I got caught up in the KEP freeze rush!

PR #185, opened 2022-03-17
Added a new flag --update-latest-green-marker to the GKE deployer. 

richackard:
Changed the build marker staging logic for the GKE Deployer so that it only gets staged to GCS after the build / test is completed successfully.

### DIFF
--- a/kubetest2-gke/deployer/build/gke_make.go
+++ b/kubetest2-gke/deployer/build/gke_make.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/kubetest2-gke/deployer/build/gke_make.go
+++ b/kubetest2-gke/deployer/build/gke_make.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


BenTheElder:
> Changed the build marker staging logic for the GKE Deployer so that it only gets staged to GCS after the build / test is completed successfully.

I'm not sure this is something we should be doing, it seems like an abuse of the build marker.

What happens if I only run --build?

If this is meant to be a marker of test passing, why not upload a distinct marker for that?

richackard:
Let's make the original marker serve its own purpose. Will add a flag to control whether we should also stage a "green version" of the build marker.

kubetest2-gke/deployer/build/gke_make.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
```
BenTheElder, line 2: this should remain the original year. I can't find anything written down but throughout the project we operate this way unless it is a generated file

richackard: Done

BenTheElder (approved):
/lgtm
/approve

kubetest2-gke/deployer/utils/util.go
```diff
@@ -0,0 +1,52 @@
...
+	if len(m) < 2 {
+		klog.Warningf("can't find the minor version of %s, defaulting to latest.txt", version)
+		fName = fmt.Sprintf("%s.txt", markerPrefix)
+	} else {
+		minor := m[1]
+		fName = fmt.Sprintf("%s-%s.txt", markerPrefix, minor)
+	}
+	pushCmd := fmt.Sprintf("gsutil -h 'Content-Type:text/plain' cp - %s/%s", stageLocation, fName)
+	cmd := exec.RawCommand(pushCmd)
+	cmd.SetStdin(strings.NewReader(version))
```
mborsz, line 46: This exports the marker without "v" prefix which is present for standard build markers (e.g. latest.txt).

Lack of it makes it impossible to use the latest-green marker as part of prow job (setting --extract=gs://.../latest-green-1.24.txt fails with trying to fetch build without "v" which doesn't exist).

Can this be unified with existing latest file format?

richackard: Valid point, will submit another PR to make it consistent. Thanks

PR #189, opened 2022-04-03
Ensure Down can run on build and test failure

### LINKED ISSUE #188: Kubetest2 doesn not down on ginkgo failures
Previously, kubetest2 would have a `defer shouldDown()` that ensured clusters were always cleaned up after run regardless of outcome given that `--down` flag was supplied.

https://github.com/kubernetes-sigs/kubetest2/commit/d9ca6c849f1e275cc5b98c1248bfeb24e72c684a seems to have removed that defer and instead only ensures down is executed on certain signals. If ginkgo returns an error, `shouldDown` will never be run.

The effect of this is rather massive resource leakage in the kOps AWS account.

We'd like to understand if this is intended behavior or if this is a bug.

/cc @ShwethaKumbla

olemarkus:
Add the down logic back into a `defer` so that resources are cleaned up on cluster up and test failures.
Fixes #188 
Fixes https://github.com/kubernetes/kops/issues/13349

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -127,6 +127,18 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		}
 	}
 
+	// ensure tearing down the cluster happens last.
+	// down should be called both when Up and Test fails to ensure resources are being cleaned up.
+	defer func() {
+		if opts.ShouldDown() {
+			// TODO(bentheelder): instead of keeping the first error, consider
+			// a multi-error type
+			if err := writer.WrapStep("Down", d.Down); err != nil && result == nil {
+				result = err
+			}
+		}
+	}()
+
 	// up a cluster
 	if opts.ShouldUp() {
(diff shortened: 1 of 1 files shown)


hakman:
This reverts part of https://github.com/kubernetes-sigs/kubetest2/pull/148.
/cc @ShwethaKumbla @amwat

olemarkus:
I see that the defer was removed out of a concern that down will be called multiple times. But since os.Exit is called, this should not be the case. os.Exit will not call deferred functions.

pkg/app/app.go
```diff
@@ -127,6 +127,18 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		}
 	}
 
+	// ensure tearing down the cluster happens last.
+	// down should be called both when Up and Test fails to ensure resources are being cleaned up.
+	defer func() {
+		if opts.ShouldDown() {
```
BenTheElder, line 133: perhaps should be `if opts.ShouldDown() { defer ....`

PR #191, opened 2022-05-10
support creating private GKE Autopilot clusters

ZhengzheYang:
Currently creating Autopilot private clusters will give an error since `--enable-ip-alias` is not supported with Autopilot. Change the command to use the ones suggested in the [official docs](https://cloud.google.com/kubernetes-engine/docs/how-to/private-clusters).


BenTheElder (approved):
/lgtm
/approve

Issue #17, opened 2020-08-05
Release


chrislovecnm:
Do we have a release or beta release cut of this?  Where can I download it from?

amwat:
Not as of now, but we plan to cut a tagged release alongside whenever some of the kubernetes CI jobs start using the kubetest2 deployers (which are under construction). 
For now you can get the latest binaries by `go get sigs.k8s.io/kubetest2@latest` and similarly for the deployers and testers.

chrislovecnm:
@amwat any idea on timeframe?  I would love an alpha release since I am having to upload a binary to gcs for one of my builds.

amwat:
We still don't have all the functionality from kubetest yet which I probably want before the first release. Most likely some time in the 1.20 time frame. 

Does go get latest work for you currently?

chrislovecnm:
Bump. We could use a release 😁

amwat:
Ack, I'd like to get in a few things and line up `v0.1.0` with the alpha milestone of https://github.com/kubernetes/enhancements/tree/master/keps/sig-testing/2464-kubetest2-ci-migration#graduation-criteria

/cc @spiffxp @BenTheElder

BenTheElder:
it might be worth trialing with something like v0.0.1 just to exercise the mechanisms intended for use and let users start using fixed tags more easily. we can have many v0.0.Z to test it out.

chrislovecnm:
Can we please have a beta or alpha release??  Pretty please :)

chrislovecnm:
Bump

amwat:
I was mainly waiting on the node tester changes in https://github.com/kubernetes-sigs/kubetest2/pull/103
there's only 1 last issue to tackle in the milestone https://github.com/kubernetes-sigs/kubetest2/milestone/1

after which I'll cut a v0.1.0

chrislovecnm:
Any updates???????  Please cut a release

amwat:
Ah yes, I've fallen a bit behind on this 😅 . All the things that we were mostly waiting on are in now. 
Will cut a release this week.

pseudomuto:
Is this still something that's going to happen? Any reason to not cut a release?

BenTheElder:
amwat moved on and this project is understaffed, I'm currently the only active approver and I have a number of other more urgent things elsewhere

GanjMonk:
[...]

PR #192, opened 2022-05-20
Set up GoReleaser to make releases for tags

pseudomuto:
Related: #17

Adding [GoReleaser](https://goreleaser.com/) to handle generating
releases for tags. This will make it so that cutting releases is as
simple as pushing a tag. The result will be a release which includes the
changelog, binaries for Linux/Windows/Darwin, as well as the checksums
for each of the archives.

To see what this looks like, I've created a dummy release on my fork.
Because this is the first release, the changelog is massive, but going
forward it should be a little more reasonable.

Sample release:
https://github.com/pseudomuto/kubetest2/releases/tag/v0.1.0

I've added some light details about how it works in RELEASE.md as well
as the one-time setup step required to make it work (A limited scope
GitHub token).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,27 @@
+---
+name: Release
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      - name: Release
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_RELEASE_TOKEN }}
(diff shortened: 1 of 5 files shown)


MushuEE:
I don't have any context on this change, was there a conversation in the slack about the need for this framework? @BenTheElder

BenTheElder (commented):
Thanks but, this isn't an approach the Kubernetes project uses to releasing, cc @kubernetes-sigs/release-engineering  ... and the problem with #17 is "nobody has time to work on this project" (I try to make time to at least review PRs, but that's honestly well beyond what I can commit to) not "running builds and uploading them to a release page takes too long".

I am focused right now on finding an approver with time commitment.

> This will make it so that cutting releases is as simple as pushing a tag.

I disagree, releases come with an expectation of a level of stability and support when we cannot guarantee any. Cutting a release is more about ensuring things are in the right state and documenting noteworthy changes, not about building and uploading binaries or adding git tags.

I don't want to start cutting releases now and give a false impression of a level of support that does not exist. This repo is getting best-effort PR reviews at this time, nothing more.

If folks would like to help https://github.com/kubernetes-sigs/kubetest2/blob/master/CONTRIBUTING.md we can work on reaching a level of trust to hand off the project.

If anyone wishes to pin to a specific commit themselves, that's low-cost to do without any changes to the repo. Versioning the repo adds maintenance expectations that we cannot support at this time.

pseudomuto:
Ok, thanks for clarifying.

PR #198, opened 2022-06-28
Switch to golang native error wrapping

saschagrunert:
We now use the native error wrapping support since `github.com/pkg/errors` is read-only.


BenTheElder:
this does mean a loss of stack traces though, and pkg/errors doesn't have much code in it.

given how little we appear to be using it to begin with though ...
/lgtm
/approve

PR #199, opened 2022-07-01
Bump golang to 1.17

olemarkus:
Would be ideal to go up to 1.18, but that involved quite a bit more work.


olemarkus:
Appreciate the notice.

PR #201, opened 2022-07-20
revert naming of FlakeAttempts

GanjMonk:
premature change of naming, ginkgo v2 is not running when kubetest2 installed from scratch

### DIFF
--- a/pkg/testers/ginkgo/ginkgo.go
+++ b/pkg/testers/ginkgo/ginkgo.go
@@ -65,7 +65,7 @@ func (t *Tester) Test() error {
 	e2eTestArgs := []string{
 		"--kubeconfig=" + t.kubeconfigPath,
 		"--kubectl-path=" + t.kubectlPath,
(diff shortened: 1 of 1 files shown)


BenTheElder:
I think we can do better for this, like detecting the ginkgo binary version or something. We *do* need to work with Kubernetes development at HEAD as well as past release branches.

cc @chendave

chendave:
> I think we can do better for this, like detecting the ginkgo binary version or something.

+1, maybe a change like this one https://github.com/kubernetes/test-infra/pull/26250


> We do need to work with Kubernetes development at HEAD as well as past release branches.

+1 too,  the support for `flakeAttempts` will eventually gone with ginkgo v2,  I am afraid this will break the test again at some time in the future.

aojea:
I have a better fix for this

BenTheElder:
thanks for the PR -- please check if antonio's solution works for you https://github.com/kubernetes-sigs/kubetest2/pull/206,

spiffxp:
/close
Closing in favor of https://github.com/kubernetes-sigs/kubetest2/pull/206

PR #206, opened 2022-07-31
ginkgo v2 backwards compatible

### LINKED ISSUE #204: Ginkgo v2 flag changes break testing older k8s versions
kOps uses kubetest2 to test all supported k8s versions. The flag changes from upgrading to ginkgo v2 (https://github.com/kubernetes-sigs/kubetest2/pull/200) impact all usage of kubetest2. Because ginkgo was only upgraded to v2 in k/k's master branch, older k8s versions still use ginkgo v1. This inconsistency results in failures to test older k8s versions.

[example with k8s 1.22](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kops/14061/pull-kops-latest-e2e-aws-k8s-1-22/1552987041279512576) from https://github.com/kubernetes/kops/pull/14061

```
	 -------------------------------------------------------------------
	|                                                                   |
	|  Ginkgo timed out waiting for all parallel nodes to report back!  |
	|                                                                   |
	 -------------------------------------------------------------------

 e2e timed out. path: ./_rundir/d40e3fa1-0f35-11ed-80a5-ea2e797f9e54
[1] flag provided but not defined: -ginkgo.flake-attempts
[1] Usage of /home/prow/go/src/k8s.io/kops/_rundir/d40e3fa1-0f35-11ed-80a5-ea2e797f9e54/e2e.test: 
...
```

Can we make kubetest2's ginkgo flag changes conditional based on the k8s version being tested? Or what is the expected path forward to use the latest kubetest2 while testing older supported k8s versions?

aojea:
some ginkgo flags and behaviors are not backwards compatible, detect the major version to configure accordingly 

Fixes #204

### DIFF
--- a/pkg/testers/ginkgo/ginkgo.go
+++ b/pkg/testers/ginkgo/ginkgo.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/kballard/go-shellquote"
 	"github.com/octago/sflags/gen/gpflag"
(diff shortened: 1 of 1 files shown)


aojea:
/assign @BenTheElder @chendave @MushuEE 

I couldn't test it, is there a way to verify it works correctly?

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -155,6 +170,29 @@ func (t *Tester) validateLocalBinaries() error {
...
+	cmd := exec.Command(t.ginkgoPath, "version")
+	lines, err := exec.OutputLines(cmd)
+	if err != nil || len(lines) != 1 {
+		return ""
+	}
+	// the output is in the format
+	// Ginkgo Version 1.14.0
+	// Ginkgo Version 2.1.4
+	parts := strings.Split(lines[0], " ")
+	if len(parts) != 3 {
```
chendave, line 186: Do you want to print some info here instead of just return an empty string?

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -155,6 +170,29 @@ func (t *Tester) validateLocalBinaries() error {
...
+		return ""
+	}
+	vers := strings.Split(parts[2], ".")
+	if len(vers) != 3 {
```
chendave, line 190: not sure if ginkgo will has a version like "2.x.x.x" or "2.x" in the future, is the major parsed from the string is enough?

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -155,6 +170,29 @@ func (t *Tester) validateLocalBinaries() error {
 	return nil
 }
 
+// ginkgoMajorVersion returns the ginkgo major version
+// empty if not found
```
chendave, line 174: empty if not found or on error.

chendave (commented):
Verified the code, this change should work for the current versions, not familiar with kubetest2 though.
Just some nits or question left, thanks for the fix, LGTM.

BenTheElder (commented):
given the severity of the breakage for some folks, and the large timezone skew, let's merge and iterate. thanks @aojea
/lgtm
/approve

PR #211, opened 2022-08-11
synchronize systemout

### LINKED ISSUE #210: Unsafe concurrent use of bytes.Buffer
https://github.com/kubernetes-sigs/kubetest2/blob/master/pkg/process/junitexec.go#L59-L61 uses the same bytes.Buffer in two output streams, which means it can be written to concurrently. That is not threadsafe.

This panics randomly with changes made to bytes.Buffer growing in go1.19.

The code should be changes to make use of the buffer threadsafe or use two separate buffers

BenTheElder:
fixes #210 

NOTE: we don't bother synchronizing `.String()` because the process is errored and dead when we're reading it.


xmcqueen:
nice
/lgtm

PR #213, opened 2022-08-23
Update REVIEWERS

MushuEE:
Update reviewers to increase bus factor on Kubetest2


OWNERS
```diff
@@ -2,9 +2,10 @@
 
 reviewers:
 - MushuEE
+- spiffxp
```
spiffxp, line 5: FWIW I'm willing to go back to approver if you're OK with it

spiffxp: I opened https://github.com/kubernetes-sigs/kubetest2/pull/215 to add me as approver

We'll need to get @richackard added as an org member

spiffxp (changes requested):
Rebase to squash and drop the `@` mentions from the commit message to get rid of the `do-not-merge` label

PR #224, opened 2022-09-02
dump cluster logs to artifacts dir instead of run dir

spiffxp:
Fallout from https://github.com/kubernetes-sigs/kubetest2/pull/183

Logs used to be uploaded because rundir was inside of ARTIFACTS, e.g. https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/ci-kubernetes-gce-conformance-latest-kubetest2/1540532050887446528/artifacts/eb7a368d-f433-11ec-9e31-9224b4edca5e/

### DIFF
--- a/kubetest2-gce/ci-tests/buildupdown.sh
+++ b/kubetest2-gce/ci-tests/buildupdown.sh
@@ -28,6 +28,17 @@ make install-deployer-gce
 # currently equivalent to /home/prow/go/src/github.com/kubernetes/cloud-provider-gcp
 K_REPO_ROOT="${REPO_ROOT}/../../kubernetes/cloud-provider-gcp"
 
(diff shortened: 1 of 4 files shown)


spiffxp:
I'm wondering if https://github.com/kubernetes/cloud-provider-gcp/pull/369/commits/84826703ed30c052de50eca706cb291ea30e4cc5 broke the gce-build-up-down job

spiffxp:
Tried manually reverting that commit in a docker run of kubekins, same error

Now my guess is the gce-build-up-down job needs to move to bazel 5.3.0 as was done for the cloud-provider-gcp jobs (e.g. https://github.com/kubernetes/test-infra/pull/27335) though I'm not sure what required it

PR #226, opened 2022-11-14
Don't shut down on control-C, if we should not shut down

justinsb:
If the user has asked us not to shut down the cluster, don't shut it
down if they control-C out of long running tests.

Allows for iterative test debugging etc.


spiffxp (approved):
/approve
/lgtm

PR #233, opened 2023-07-21
Add an env for `timeout` for ginkgo tests

chendave:
By default, it only allows the ginkgo test to run [45 minutes](https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/remote/remote.go#L34), then the test will abort.

But it's not enough to run slow tests, in order to run slow test, we should provide a way to define the `timeout` if the CI job is run by `kubetest2`

-----

You can find this in the [log](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/118650/pull-kubernetes-node-arm64-ubuntu-serial-gce/1672445937110224896/artifacts/build-log.txt) where slow tests are executed.

> [sig-node] Restart [Serial] [Slow] [Disruptive] Container Runtime Network should recover from ip leak
...
Interrupted by User
First interrupt received; Ginkgo will run any cleanup and reporting nodes but will skip all remaining specs. Interrupt again to skip cleanup.
/bin/bash: line 1: 2134 Killed timeout -k 30s 2700.000000s ./ginkgo ...

### DIFF
--- a/pkg/testers/node/node.go
+++ b/pkg/testers/node/node.go
@@ -47,28 +47,29 @@ const (
 )
 
 type Tester struct {
-	RepoRoot                       string `desc:"Absolute path to kubernetes repository root."`
-	GCPProject                     string `desc:"GCP Project to create VMs in. If unset, the deployer will attempt to get a project from boskos."`
-	GCPZone                        string `desc:"GCP Zone to create VMs in."`
-	SkipRegex                      string `desc:"Regular expression of jobs to skip."`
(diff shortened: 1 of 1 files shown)


chendave:
/cc @dims
who knows the context of this.

chendave:
Looking for something to approve this, 

/assign @michelle192837 @MushuEE 

Can you pls help to review? thanks!

michelle192837 (approved):

PR #240, opened 2023-08-18
configure kubetest2 to build arm64 e2e tests

upodroid:
/cc @dims @tzneal 

In theory, all we need to do to run arm e2e tests on GCP is to compile arm binaries, create VMs with arm machine types and use the arm images


ruquanzhao:
It seems the CI failed because some hardcode in k/k shell scripts.
I raise a PR about env KUBE_BUILD_PLATFROMS in https://github.com/kubernetes/kubernetes/pull/120100

neolit123:
https://github.com/kubernetes/kubernetes/pull/120144#issuecomment-1707041256

upodroid:
@michelle192837 Can you approve this please? we can ignore the failing aws test as it is unrelated to this change.

michelle192837 (approved):

PR #241, opened 2023-09-02
Support for --cloud-provider and --featuregates

dims:
Needed for https://github.com/kubernetes/kubernetes/issues/120367

we have to set both the cloud-provider (to `gce`) and the disable feature gates (for in-tree cloud provider and in-tree kubelet cred provider)


dims:
@jbpratt did you mean to `/approve` as well? :)

jbpratt:
/approve

@dims :+1:

PR #244, opened 2023-09-26
[cl2] pass through a couple of prometheus parameters

dims:
Needed for use by scalability jobs - https://github.com/kubernetes/kops/pull/15963


dims:
/skip

https://testgrid.k8s.io/sig-testing-canaries#pull-kubetest2-gce-legacy-build-up-down&width=20 shows that the new job is still under construction

PR #245, opened 2023-09-26
Fix legacy updown cluster tests

upodroid:
I wanted to amend kube-up clusters to pick the latest image in an image family instead of using a hardcoded image. 

Turns out it is a messy change: https://github.com/kubernetes/kubernetes/pull/120144

/cc @jbpratt @michelle192837


upodroid:
/retest

known flakes

PR #250, opened 2023-11-11
Add --use-binaries-from-path to ginkgo tester

cartermckinnon:
Adds an option to the ginkgo tester to locate the `ginkgo`, `e2e.test`, and `kubectl` binaries from the system's PATH.

### DIFF
--- a/pkg/testers/ginkgo/ginkgo.go
+++ b/pkg/testers/ginkgo/ginkgo.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	stdexec "os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


pkg/testers/ginkgo/ginkgo.go
```diff
@@ -172,6 +177,26 @@ func (t *Tester) validateLocalBinaries() error {
...
+		}
+		klog.V(2).Infof("found existing %s at %s", binary, path)
+		switch binary {
+		case "e2e.test":
+			t.e2eTestPath = path
+		case "ginkgo":
+			t.ginkgoPath = path
+		case "kubectl":
+			t.kubectlPath = path
+		}
```
cartermckinnon, line 195: This is kind of gross, I could use a `map[string]string` to store the binary paths instead of having a separate field in the `Tester` for them.

dims (approved):
/approve
/lgtm

/assign @michelle192837 @xmcqueen @jbpratt

michelle192837 (approved):

Issue #239, opened 2023-08-16
Is gsutil package required to run kubetest2?


pbxqdown:
I'm running with `kind` deployer on my local machine(`KUBE_GIT_VERSION=v1.27.0 kubetest2 kind --up --test=ginkgo -- --focus-regex='\[Conformance\]'`) but get following error:
```
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Have a question, bug, or feature request? Let us know! https://kind.sigs.k8s.io/#community 🙂
I0816 00:09:13.279804   12136 ginkgo.go:139] Using kubeconfig at /home/qian/.kube/config
F0816 00:09:13.280042   12136 ginkgo.go:255] failed to run ginkgo tester: failed to get ginkgo test package from published releases: failed to get latest release name: exec: "gsutil": executable file not found in $PATH
Error: exit status 255
```

Since kind is used for testing, I think kubetest2 shouldn't depend on any cloud-provider?

dims:
@pbxqdown if you have `kubectl` in path, it will not try to get a fresh one from our published artifacts

pbxqdown:
@dims thanks, is there a way to resolve the issue or doc to follow the process? In a k8s developer's environment `kubectl` is usually included.

PR #251, opened 2023-11-21
Download binaries from dl.k8s.io instead of a bucket

upodroid:
/cc @ameukam @BenTheElder @dims

Fixes: https://github.com/kubernetes-sigs/kubetest2/issues/239
Part of: https://github.com/kubernetes/test-infra/issues/18789 and https://github.com/kubernetes/k8s.io/issues/2396

kops is another large project that fetches from the kubernetes-release/k8s-release-dev buckets but it has a custom implementation at https://github.com/kubernetes/kops/tree/master/tests/e2e/pkg/tester that I'll fix separately.

Lets merge this as soon as code freeze is over and bump the kubekins-e2e image. We should start seeing jobs failing and we can fix them

Improvements:
- kubetest2 can now fetch kube binaries from the internet instead of building it from source
- ginkgo can fetch kubectl and test binaries from dl.k8s.io or storage.googleapis.com/kubernetes-release


kubetest2-gce/ci-tests/buildupdown-legacy.sh
```diff
@@ -27,8 +27,6 @@ cd "${GOPATH}/src/k8s.io/kubernetes"
 # kubetest2 against k/k
 kubetest2 gce \
     -v=2 \
-    --repo-root=. \
```
upodroid, line 30: I'm going to revert the change to this file and create a prowjob for the new script in this folder that specifically tests prebuilt k/k scenarios.

pkg/testers/ginkgo/ginkgo.go
```diff
@@ -39,13 +38,12 @@ import (
...
 
 type Tester struct {
-	FlakeAttempts       int           `desc:"Make up to this many attempts to run each spec."`
 	GinkgoArgs          string        `desc:"Additional arguments supported by the ginkgo binary."`
 	Parallel            int           `desc:"Run this many tests in parallel at once."`
 	SkipRegex           string        `desc:"Regular expression of jobs to skip."`
 	FocusRegex          string        `desc:"Regular expression of jobs to focus on."`
+	TestPackageURL      string        `desc:"The url to download a kubernetes test package from."`
 	TestPackageVersion  string        `desc:"The ginkgo tester uses a test package made during the kubernetes build. The tester downloads this test package from one of the release tars published to the Release bucket. Defaults to latest. visit https://kubernetes.io/releases/ to find release names. Example: v1.20.0-alpha.0"`
-	TestPackageBucket   string        `desc:"The bucket which release tars will be downloaded from to acquire the test package. Defaults to the main kubernetes project bucket."`
```
rifelpet, line 48: This is causing test failures: https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-aws-selinux/1744533959024316416

can we finish migrating all jobs to use the new flag, or add this back as a deprecated flag?

PR #252, opened 2023-12-05
export all deployer struct

errordeveloper:
this will ensure that each deployer can be easily called as a library, which GKE deployer already allows (kubernetes-sigs/kubetest2#135)

### DIFF
--- a/kubetest2-gce/deployer/build.go
+++ b/kubetest2-gce/deployer/build.go
@@ -28,7 +28,7 @@ import (
 	"sigs.k8s.io/kubetest2/pkg/exec"
 )
 
-func (d *deployer) Build() error {
+func (d *Deployer) Build() error {
 	klog.V(1).Info("GCE deployer starting Build()")
 
 	if err := d.init(); err != nil {
@@ -91,7 +91,7 @@ func (d *deployer) Build() error {
 	return nil
 }
 
-func (d *deployer) setRepoPathIfNotSet() error {
+func (d *Deployer) setRepoPathIfNotSet() error {
 	if d.RepoRoot != "" {
 		return nil
 	}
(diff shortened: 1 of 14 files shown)


upodroid:
Hi

Thank you for opening this PR. The deployers are effectively a finished program that shouldn't be called as libraries. If there is a missing feature, we recommend that you open a PR here and add it.

We have external implementations listed [here](https://github.com/kubernetes-sigs/kubetest2#external-implementations) that use pkg/* to create their own deployers.

What problem are you trying to solve?

errordeveloper:
@upodroid thanks for your comment. I thought this change would be reasonable given GKE deployer is public (see #135).

I'd like to use kubetest2 in a project and will likely have a custom tester. I'd rather avoid creating a bash wrapper or a go program that uses exec, I don't see what would be the issue in exporting these structs, given GKE one is already is already exported. If there was a major objection, I suppose you should put the package under `internal/`, which would give a stronger signal that something is not intended to be shared outside the repo.

> We have external implementations listed [here](https://github.com/kubernetes-sigs/kubetest2#external-implementations) that use pkg/* to create their own deployers.

What I want to do is use GKE and kind deployers, at the moment I can use the GKE one, but not kind. I'd rather not fork/rewrite same code.

upodroid:
> I'd like to use kubetest2 in a project and will likely have a custom tester. I'd rather avoid creating a bash wrapper or a go program that uses exec, I don't see what would be the issue in exporting these structs, given GKE one is already is already exported.

kubetest2 should be used like this. https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gce/ci-tests/buildupdown-legacy.sh or https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-tester-node/ci-tests/gce-test.sh It should be the only program you call to run the entire process of creating a cluster, testing it and tearing it down.

The GKE one was exported for another project I work on and it wasn't a good idea. https://github.com/knative/hack/pull/230 we embedded that deployer in another program and it was painful to work with and I got rid of it.

PR #253, opened 2023-12-17
Add DeployerWithInit interface

cartermckinnon:
Adds a new interface that deployers can implement, `DeployerWithInit`, which allows a deployer to define initialization behavior. The deployer's `Init` function will be called prior to build/up/down/etc.

The initialization of a deployer may depend on its options, so cannot be done in `NewDeployer` (called when constructing the CLI).


upodroid (approved):
LGTM

The failing job requires a file being added in #251 

/assign @michelle192837 @jbpratt

PR #255, opened 2023-12-20
Add DeployerWithFinish interface

cartermckinnon:
Similar to #253, adding the ability to define a finalizer function that is called regardless of whether `--down` is used. This can be used to e.g. emit metrics, a lifecycle event, etc.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -125,6 +125,12 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		if err := junitRunner.Close(); err != nil && result == nil {
 			result = err
 		}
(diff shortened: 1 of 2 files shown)


cartermckinnon:
@dims do you have an opinion on this one? It would give me a more natural place to emit metrics here: https://github.com/aws/aws-k8s-tester/pull/404

I could also change `DeployerWithInit` to `DeployerWithLifecycle` and roll this into that interface. It was added a few weeks ago so I'm not really concerned about breakage.

upodroid (approved):
LGTM

/assign @jbpratt @michelle192837

dims:
/approve
/lgtm

works for me @cartermckinnon

PR #260, opened 2024-02-15
build darwin binaries and fix cloudbuild config

upodroid:
go 1.22 seems to be compiling arm64 properly instead of 1.21 :(


upodroid:
bump

/cc @jbpratt

PR #261, opened 2024-03-08
Make it obvious where test tar ball is downloaded from

dims:
Took me some time to figure out that the test tar ball was not being downloaded from where i expected it to be! Let's please log an explicit message.


TerryHowe (approved):
/lgtm

Issue #263, opened 2024-04-12
Problem with passing the provider to the tests


jbtk:
Hi!

I am trying to run e2e autoscaler tests on a cluster on GCE.

When I run: 
kubetest2 gce -v 2   --repo-root <path>   --gcp-project <name>   --legacy-mode --build --up --down   --test=ginkgo   --   --focus-regex='\[Feature:ClusterSizeAutoscalingScaleUp\]|\[Feature:ClusterSizeAutoscalingScaleDown\]|\[Feature:InitialResources\]' --ginkgo-args="-v"

Most of the tests are skipped because it seems that the tester does not have information which provider was used to provision the cluster for the test. I get information:
[SKIPPED] Only supported for providers [gce gke] (not skeleton)

I tried looking for "skeleton" and I see it in only one place in kubetest2 repo - in the clusterloader2 which seems to be only used by the clusterloader tester so I am not sure why this is a problem with ginkgo - I am probably missing some call stack as in the code they do not seem to depend on each other. 

Am I doing something wrong? Is it a known issue?

jbtk:
Testing some more:
- the clusterloader2 is not relevant
- the provider param is missing in the ginkgo runner and since it is not passed the test_context.go uses skeleton as a fallback

jbtk:
I have read the discussion in https://github.com/kubernetes/kubernetes/issues/70194 and I am still not sure what is the up to date plan going forward, but it seems that the best option would be to make the tests that I am running provider agnostic and remove dependency on the provider. I will look into that, but I am still not sure whether it is expected for kubetest2 not to support passing provider or is it a missing functionality. Would be happy to learn that. I could contribute a fix here, but would like to make sure that it is indeed wanted.

jbtk:
It seems that even after hacking (hardcoding locally to check whether it would work) and passing the provider the test will not run as they require other context args that are not populated (as defined in https://github.com/kubernetes/kubernetes/blob/master/test/e2e/framework/test_context.go#L347). Would it make sense for deployer to prepare set of provider specific e2e test args that could be read from the rundir similarly to the kubeconfig?

jbtk:
I found a workaround to pass the provider manually in the test args. Example:
[...]

PR #264, opened 2024-04-25
Instead of overriding the env variables add them to the environment.

jbtk:
Before this change if needed to pass additional env variable to the ginkgo tester it would remove other variables. With this change it adds variables specified in the commandline.

### DIFF
--- a/pkg/testers/ginkgo/ginkgo.go
+++ b/pkg/testers/ginkgo/ginkgo.go
@@ -99,7 +99,9 @@ func (t *Tester) Test() error {
 
 	klog.V(0).Infof("Running ginkgo test as %s %+v", t.ginkgoPath, ginkgoArgs)
 	cmd := exec.Command(t.ginkgoPath, ginkgoArgs...)
-	cmd.SetEnv(t.Env...)
+	env := os.Environ()
+	env = append(env, t.Env...)
+	cmd.SetEnv(env...)
 	exec.InheritOutput(cmd)
 	return cmd.Run()
 }


jbtk:
It seems that the test is setting env variable to use bazel 5.3.0 and then something else expects that bazel version is >= 5.4.0. 

- the version is set here: https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gce/ci-tests/buildupdown.sh#L31 - is this actually needed? Can it be bumped up to 5.4? It is likely that this is caused by this change- as it prevents the env from being fully overwritten.

jbtk:
I pushed a new version, but it was only caused by some changes in my local repo. I moved the change of the version to a separate PR (https://github.com/kubernetes-sigs/kubetest2/pull/265). Please let it test and approve if it is ok. Once this is done I will update this version so that the test passes.

BenTheElder:
> Before this change if needed to pass additional env variable to the ginkgo tester it would remove other variables. With this change it adds variables specified in the commandline.

Those values should be added as flags to the tester ..?

The not passing through env is by design. It makes tests more reproducible.

jbtk:
The cl came from the fact that I struggled to run tests in legacy mode (see issue for details: https://github.com/kubernetes-sigs/kubetest2/issues/263). 

It seems that kubetest2 sets a various env variables and therefore passing one additional and clearing them breaks the test (at least this is what happens in the legacy mode). I am not sure whether expecting person who wants to add/update one env variable to know how to specify all of them is a good solution (though I might be missing something - as you see in the issue running tests was a journey for me because I do not know the ecosystem well enough). From my testing it seemed that passing empty does not clear previously existing env and passing a single one does clear them and sets only the defined env variable. This is confirmed in the Env field of the Cmd: https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/os/exec/exec.go;l=163. So currently we are actually passing all the env vars unless user specifies env flags and then we only set what is specified (and clear what would be normally passed otherwise).

jbtk:
/remove-lifecycle stale

Since I explained why I proposed this PR and did not get any response I am restarting the clock. If I do not get any answer in 30d I will just close this pr.

TerryHowe:
Since overriding the env is by design, it seems like the only option to make tests deterministic would be add an option or configuration with overrides. Otherwise close this maybe.

PR #265, opened 2024-05-09
Update the bazel version. 

jbtk:
Otherwise the test fails with "Current Bazel version is 5.3.0, expected at least 5.4.0".


kubetest2-gce/ci-tests/buildupdown.sh
```diff
@@ -29,7 +29,7 @@ make install-tester-ginkgo
 REPO_ROOT="${GOPATH}"/src/k8s.io/cloud-provider-gcp;
 
 # TODO(spiffxp): remove this when cloudprovider-gcp has a .bazelversion file
-export USE_BAZEL_VERSION=5.3.0
+export USE_BAZEL_VERSION=5.4.0
```
BenTheElder, line 32: we should really control this from cloudprovider-gcp

BenTheElder: cc @seans3

jbtk:
Can we check at least whether with this change the test passes? This test is the only failing in my other PR and since I was able to reproduce without my change I guess that it would fail on current master as well (not sure, was not able to find any CI that would run these tests).

TerryHowe (approved):
I think this is g2g
/lgtm

https://github.com/kubernetes/cloud-provider-gcp/blob/master/.bazelversion

BenTheElder (commented):
/lgtm
/approve

upodroid (approved):

BenTheElder:
> mushuee

I don't think @MushuEE is doing OSS development these days and we should probably update OWNERS. (spiffxp is not either)

maybe @aojea wants more to do ;-)

aojea:
> maybe @aojea wants more to do ;-)

https://github.com/kubernetes-sigs/kubetest2/pull/267

PR #266, opened 2024-07-02
Support "--base-image" in kind deployer's build phase.

everpeace:
In kind, container runtimes(high level and low level) are installed in [Base Image](https://kind.sigs.k8s.io/docs/design/base-image/).

Sometimes, E2E which requires custom-built container runtimes(contained, CRI-O, crun, etc.) needs to use custom base image when building kind node image.

But, because current `kubetest2-kind` does not support specifying the base image in `build` phase, kubetest2 users need to build the node image separately and then run kind deployer like this:

```shell
$ kind build node-image --base-image=... --image=...
$ kubetest2 kind --up --image=... --test=...
```

Thus, this PR just adds `--base-image` in kind deployer to make it in just one line:

```shell
$ kubetest2 kind --build --up --base-image=... --test=...
```

### DIFF
--- a/kubetest2-kind/deployer/build.go
+++ b/kubetest2-kind/deployer/build.go
@@ -35,6 +35,9 @@ func (d *deployer) Build() error {
 	if d.KubeRoot != "" {
 		args = append(args, "--kube-root", d.KubeRoot)
 	}
+	if d.BaseImage != "" {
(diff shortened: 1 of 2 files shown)


upodroid (approved):
LGTM, Thanks

Issue #271, opened 2024-10-14
Logs are obtained after the cluster is up and before the tests run


aojea:
As a consequence there is no possibility of debugging test failures, IIRC the logs should be obtained before we are tearing down the cluster, but when this option was added , it got added to the `Up()` hook https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-kubernetes-gce-conformance-latest-kubetest2/1845682784069226496

https://github.com/kubernetes-sigs/kubetest2/commit/fb21a08299d15675caf10f2cf9966faf6933ce94

It seems the AWS one logs correctly

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/ci-cloud-provider-aws-e2e-kubetest2/1845768599910748160

aojea:
Suggestions from @soltysh seems the best one, moving the dump of the logs to the `Down()` hook

https://github.com/kubernetes-sigs/kubetest2/blob/master/kubetest2-gce/deployer/down.go#L28

PR #272, opened 2024-10-16
Dump cluster logs before bringing the cluster down

soltysh:
Fixes https://github.com/kubernetes-sigs/kubetest2/issues/271

/assign @BenTheElder


BenTheElder (commented):
TODO: we should probably dump cluster logs after test() even if we don't down() but this seems like a reasonable stopgap

/lgtm

aojea:
/close

I inadvertently slipped this one and merged other similar to this, just replacing the defer to avoid dumping logs twice

https://github.com/kubernetes-sigs/kubetest2/pull/273/files

PR #273, opened 2024-11-02
kubetest-gce: dumpcluster logs after it run

aojea:
Only dump cluster logs on Up if the cluster is not able to created, otherwise logs them before tearing down the cluster

### DIFF
--- a/kubetest2-gce/deployer/down.go
+++ b/kubetest2-gce/deployer/down.go
@@ -28,6 +28,10 @@ import (
 func (d *deployer) Down() error {
 	klog.V(1).Info("GCE deployer starting Down()")
 
+	if err := d.DumpClusterLogs(); err != nil {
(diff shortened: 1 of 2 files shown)


aojea:
/assign @BenTheElder @ameukam @upodroid 

fixes this problem https://github.com/kubernetes/kubernetes/pull/128508

aojea:
@carlory please LGTM once CI is green, so we can investigate the flakes on the kubetest jobs, right now we can not see the logs because they are not dumped after the tests finished running

aojea:
now is correct, 

> PASS

Ginkgo ran 1 suite in 30.020464496s
Test Suite Passed
I1102 09:51:25.438492   14241 down.go:29] GCE deployer starting Down()
I1102 09:51:25.438525   14241 dumplogs.go:30] GCE deployer starting DumpClusterLogs()
I1102 09:51:25.438533   14241 dumplogs.go:36] making logs directory

it dumps the logs after the test have run

PR #274, opened 2024-11-02
Dump cluster logs when necessary

aojea:
We need the cluster logs in two situations:
- when the cluster fails to be created
- after running the tests, before we start tearing down the cluster


ameukam:
Looks like it's related to https://github.com/kubernetes-sigs/kubetest2/pull/272 ?

PR #282, opened 2024-12-09
update dependencies in hack/tools

mkumatag:
- updated the dependencies in the go.mod file
- Fixed the following lints exposed via the latest linter
```shell
WARN The linter 'exportloopref' is deprecated (since v1.60.2) due to: Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar.
pkg/app/cmd.go:103:22: printf: non-constant format string in call to fmt.Errorf (govet)
			return fmt.Errorf(stderr.String())
			                  ^
pkg/testers/clusterloader2/cl2.go:73:7: S1009: should omit nil check; len() for []string is defined as zero (gosimple)
			if s.TestConfigs != nil && len(s.TestConfigs) > 0 {
			   ^
pkg/testers/clusterloader2/cl2.go:76:7: S1009: should omit nil check; len() for []string is defined as zero (gosimple)
			if s.TestOverrides != nil && len(s.TestOverrides) > 0 {
			   ^
pkg/boskos/boskos.go:76:107: redefines-builtin-id: redefinition of the built-in function close (revive)
func startBoskosHeartbeat(boskosClient *client.Client, resource *common.Resource, interval time.Duration, close chan struct{}) {
```


upodroid (approved):
LGTM

Issue #285, opened 2024-12-12
missing --label-filter 


elieser1101:
kubetest2 does not handle the `--label-filter` and it stops jobs like the following to work properly

- https://testgrid.k8s.io/sig-node-cri-o#pr-node-kubelet-crio-cgrpv1-dra-kubetest2
- https://testgrid.k8s.io/sig-node-cri-o#pr-node-kubelet-crio-cgrpv2-dra-kubetest2

We want to avoid falling back to `focus-regex`, disscused in https://github.com/kubernetes/test-infra/pull/33934

The failing jobs are part of https://github.com/kubernetes/test-infra/issues/32567

PR #286, opened 2024-12-13
add ginkgo-flags to node tester

elieser1101:
We need to pass ginkgo flags to enable the use of `--label-filter` or any other flags.

needs https://github.com/kubernetes/kubernetes/pull/129215 to work properly
fixes https://github.com/kubernetes-sigs/kubetest2/issues/285

### DIFF
--- a/pkg/testers/node/node.go
+++ b/pkg/testers/node/node.go
@@ -53,6 +53,7 @@ type Tester struct {
 	SkipRegex                      string        `desc:"Regular expression of jobs to skip."`
 	FocusRegex                     string        `desc:"Regular expression of jobs to focus on."`
 	TestArgs                       string        `desc:"A space-separated list of arguments to pass to node e2e test."`
+	GinkgoFlags                    string        `desc:"A space-separated list of arguments to pass to ginkgo."`
 	BoskosAcquireTimeoutSeconds    int           `desc:"How long (in seconds) to hang on a request to Boskos to acquire a resource before erroring."`
 	BoskosHeartbeatIntervalSeconds int           `desc:"How often (in seconds) to send a heartbeat to Boskos to hold the acquired resource. 0 means no heartbeat."`
 	BoskosLocation                 string        `desc:"If set, manually specifies the location of the boskos server. If unset and boskos is needed"`
@@ -252,6 +253,7 @@ func (t *Tester) constructArgs() []string {
 		// https://github.com/kubernetes/kubernetes/blob/96be00df69390ed41b8ec22facc43bcbb9c88aae/hack/make-rules/test-e2e-node.sh#L113
 		"ZONE=" + t.GCPZone,
 		"TEST_ARGS=" + t.TestArgs,
+		"GINKGO_FLAGS= " + t.GinkgoFlags,
 		"NODE_ENV= " + t.NodeEnv,
 		"DELETE_INSTANCES=" + strconv.FormatBool(t.DeleteInstances),
 		"PARALLELISM=" + strconv.Itoa(t.Parallelism),


pkg/testers/node/node.go
```diff
@@ -53,6 +53,7 @@ type Tester struct {
 	SkipRegex                      string        `desc:"Regular expression of jobs to skip."`
 	FocusRegex                     string        `desc:"Regular expression of jobs to focus on."`
 	TestArgs                       string        `desc:"A space-separated list of arguments to pass to node e2e test."`
+	GinkgoFlags                    string        `desc:"A space-separated list of arguments to pass to ginkgo."`
```
pohly, line 56: Putting arbitrary, potentially space separated flags into the value of another flag leads to an escaping nightmare. It is a useful escape hatch and we should keep it, but I'd prefer to add a `LabelFilter` above, similar to the existing `FocusRegex/SkipRegex`.

pohly: The end goal is that in a Prow job we can put the entire label filter into a single entry in argv of the command that gets invoked.

pohly: https://github.com/kubernetes/test-infra/pull/33948/files passes `--timeout` via `--ginkgo-flags`. That seems common enough that it might also deserves its own, separate flag.

Or perhaps not: it's a single word that fits into `--ginkgo-flags` without causing escaping problems. Okay, let's not add a separate flag. Thinking aloud here! :hot_face:

aojea: We are not going to add flags to allow to pass anything , is against the concept of the tool https://github.com/kubernetes-sigs/kubetest2?tab=readme-ov-file#concepts

Please, add `LabelFilter` as Patrick suggests

elieser1101: Sure will remove the `GinkgoFlags` and add only the the `LabelFilter` makes sense for me

However I dind’t follow what you mean by its against the concepts?

What would happen to other ginkgoflags?should be added each one as kubetest2 flags?

aojea: I gave you the wrong link sorry, I can't find it now, but there was one place mentioning the flags problem with kubetest1  that ended being unmanageable , specially since you can end with multiple configurations that may be invalid ... if we add a "ginkgo-flags" and users can add any option there is no possibility to validate the inputs and know if the config is valid ... if you pass flags directly you don't need kubetest at all, just run ginkgo manually that is also possible

pohly: What kind of validation is kubetest2 going to do for Ginkgo flags? I don't find that quite convincing.

If we don't add `GinkgoFlags`, then let's add `(Ginko?)Timeout` if there isn't already some other way of specifying it.

pohly: And `(Ginkgo?)Parallel` for `ginkgo -p`. That could be necessary for some jobs.

We might not need `(Ginkgo)V` for `ginkgo -v`, but who knows?

aojea: @BenTheElder may remember better but adding flags just in case is what caused kube test1 to be reborn as kubetest2 just to add only the necessary things, what's the point in passing all flags directly, then just use the bash script

aojea: I'm sorry because I'm replying from my mobile and my answers are incomplete and more confused, we have a clear demand that is ginkgo.filter `LabelFilter` so let's add this here and unblock this situation.

If there are new needs for new flags let's address those in a separate PR, but not based in hypothetical cases.

If we need to reconsider the concept and design of kubetest2 then let's discuss this is sig testing forum

pohly: Agreed, let's add `LabelFilter` and not get into a design discussion here.

aojea:
@elieser1101 can you please change the PR to narrow it to only the LabelFilter argument?

pkg/testers/node/node.go
elieser1101: Makes sense, will just include the `LabelFilter`

I found this document [why kubetest2](https://docs.google.com/document/d/1Dc7xg9lq4cxdDuz20YZjunuL5eju2WIXwO32praC5hs/edit#heading=h.ppb3an7ey1of) is this what you where trying to share? @aojea how can I get access to it?

pkg/testers/node/node.go
```diff
@@ -267,6 +268,7 @@ func (t *Tester) constructArgs() []string {
 		"USE_DOCKERIZED_BUILD=" + strconv.FormatBool(t.UseDockerizedBuild),
 		"TARGET_BUILD_ARCH=" + t.TargetBuildArch,
 		"TIMEOUT=" + t.Timeout.String(),
+		"LABEL_FILTER= " + t.LabelFilter,
```
aojea, line 271:
```suggestion
		"LABEL_FILTER=" + t.LabelFilter,
```

you have a whitespace there

aojea:
/ok-to-test
/lgtm
/approve

please follow up on https://github.com/kubernetes/kubernetes/pull/129215#issuecomment-2545520472

pkg/testers/node/node.go
pohly: You need to join https://groups.google.com/g/kubernetes-sig-testing.

elieser1101: thanks, can see it now

PR #287, opened 2025-01-28
Replace opencensus vanity url to avoid expired cert

mselim00:
Adds a go.mod directive then `go mod tidy`

This repo has a dependency on [go.opencensus.io](https://github.com/census-instrumentation/opencensus-go), which has been archived since July, 2023. The certificates for that vanity URL have expired, causing downstream CI failures (and general build issues).

This is intended as a temporary fix until the dependencies here that require it can fully migrate off.

See
- https://github.com/aws-observability/aws-otel-collector/issues/2946 
- https://github.com/googleapis/google-cloud-go/issues/11481#issuecomment-2607790138

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,8 @@ go 1.22.8
 
 toolchain go1.22.10
 
+replace go.opencensus.io => github.com/census-instrumentation/opencensus-go v0.24.0
+
 require (
 	github.com/blang/semver/v4 v4.0.0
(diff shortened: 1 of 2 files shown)


cartermckinnon:
/ok-to-test

normally want to avoid `replace`, but this particular module is dead and I wouldn't expect future releases.

cartermckinnon:
/close

`replace` will break `go install` :(

PR #333, opened 2026-03-17
Dump config maps and merge metadata sources on top of DumpClusterLogs…

vovakrk:
… since it works well with both zonal and regional clusters

### DIFF
--- a/kubetest2-gke/deployer/dumplogs.go
+++ b/kubetest2-gke/deployer/dumplogs.go
@@ -44,8 +44,6 @@ type configMap struct {
 //
 // TODO(RonWeber): This whole path is really gross, but this seemed
 // the least gross hack to get this done.
-//
-// TODO(RonWeber): Make this work with multizonal and regional clusters.
 func (d *Deployer) DumpClusterLogs() error {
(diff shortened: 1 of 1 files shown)


kubetest2-gke/deployer/dumplogs.go
```diff
@@ -47,6 +47,14 @@ type configMap struct {
...
+		log.Printf("Failed to dump configmaps: %v\n", err)
+	}
+
+	if err := d.mergeMetadata(); err != nil {
+		log.Printf("Failed to merge metadata: %v\n", err)
+	}
+
 	if len(d.Zones) <= 0 {
 		return fmt.Errorf("DumpClusterLogs is currently only supported for zonal clusters")
 	}
```
Qqkyu: Do we need the `TODO` above the function and this check if it works with regional clusters, too?

Qqkyu (commented):
/ok-to-test

kubetest2-gke/deployer/dumplogs.go
vovakrk: I moved `TODO` to the corresponding code block (several lines below)

PR #335, opened 2026-04-16
Add support for mTLS endpoint selection in GKE deployer based on gcloud configuration

apendua:
With the introduction of dedicated `mTLS` endpoints in `gcloud`

https://docs.cloud.google.com/access-context-manager/docs/understand-mtls

the hardcoded list of container API endpoints is no longer valid if client has opted in for client certificate based authentication.

This change modifies the logic of API endpoint selection by detecting wether mTLS endpoint should be used or not.

### DIFF
--- a/kubetest2-gke/deployer/commandutils.go
+++ b/kubetest2-gke/deployer/commandutils.go
@@ -31,20 +31,35 @@ import (
 	"sigs.k8s.io/kubetest2/pkg/exec"
 )
 
+var envToEndpoint = map[string]string{
+	"test":     "https://test-container.sandbox.googleapis.com/",
+	"staging":  "https://staging-container.sandbox.googleapis.com/",
+	"staging2": "https://staging2-container.sandbox.googleapis.com/",
+	"prod":     "https://container.googleapis.com/",
+}
+
+// See: https://docs.cloud.google.com/access-context-manager/docs/understand-mtls
+var envToMTLSEndpoint = map[string]string{
+	"test":     "https://test-container.mtls.sandbox.googleapis.com/",
+	"staging":  "https://staging-container.mtls.sandbox.googleapis.com/",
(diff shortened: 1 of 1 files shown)


Qqkyu (commented):
/ok-to-test

kubetest2-gke/deployer/commandutils.go
```diff
@@ -84,6 +99,15 @@ func (d *Deployer) PrepareGcpIfNeeded(projectID string) error {
 	return nil
 }
 
+func (d *Deployer) isUsingClientCertificate() (bool, error) {
+	fmt.Print("Checking if using client certificate... ")
```
Qqkyu: I think we're using `klog` everywhere, so depending on whether you think this is verbose it should probably be:

```
klog.V(1).Info("Checking if using client certificate...")
```

Or if it should be printed only when user specifies `--v 2`:

```
klog.V(2).Info("Checking if using client certificate...")
```

apendua: Fixed. I used `V(1)` for consistency with:
```
klog.V(1).Info("Checking existing of GCP ssh keys...")
```

Qqkyu:
@upodroid could you ptal and merge this if possible? The `pull-kubetest2-gce-build-up-down` is not failing because of this PR. It has failed also on 04-02 ([testgrid](https://testgrid.k8s.io/sig-testing-canaries#pull-kubetest2-gce-build-up-down)) with the same error: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/directory/pull-kubetest2-gce-build-up-down/2039684377314594816

upodroid (approved):

apendua:
@upodroid Do you think we are good to merge? The failing test seems to be unrelated.

PR #337, opened 2026-05-20
Adding tester for kube-burner

vishnuchalla:
### Description
Adding kube-burner tester for the purpose of CI. As per the plan [here](https://docs.google.com/document/d/1pNeAwIZhJQTVtYYIG-efI8xor96Q1--eSgTeXlj6cuA/edit?tab=t.0#heading=h.f8n3iqwtmmra)

### Testing
Will be testing this in integration with https://github.com/kubernetes/kops/pull/18371

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ See individual READMEs for more information
 - [`kubetest2-tester-clusterloader2`](/kubetest2-tester-clusterloader2)  - use clusterloader2
 - [`kubetest2-tester-exec`](/kubetest2-tester-exec) - exec a given command with the given args / flags
 - [`kubetest2-tester-ginkgo`](/kubetest2-tester-ginkgo) - runs e2e tests from `kubernetes/kubernetes`
+- [`kubetest2-tester-kube-burner`](/kubetest2-tester-kube-burner) - runs scalability workloads using [kube-burner](https://github.com/kube-burner/kube-burner)
 - [`kubetest2-tester-node`](/kubetest2-tester-node) - runs node e2e tests from `kubernetes/kubernetes`
 
 ## External Implementations
--- a/kubetest2-tester-kube-burner/main.go
+++ b/kubetest2-tester-kube-burner/main.go
@@ -1,0 +1,25 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


serathius:
> Adding kube-burner tester for the purpose of CI. As per the plan [here](https://docs.google.com/document/d/1pNeAwIZhJQTVtYYIG-efI8xor96Q1--eSgTeXlj6cuA/edit?tab=t.0#heading=h.f8n3iqwtmmra)

Was the plan presented to SIG-scalability and approved? I don't see any reviewers or comments.

/cc @wojtek-t

pohly:
Why does this tester need to be added and maintained here? Could it be maintained by those who use it in some other repo instead of the owners of this repo (SIG Testing)?

vishnuchalla:
> Why does this tester need to be added and maintained here? Could it be maintained by those who use it in some other repo instead of the owners of this repo (SIG Testing)?

Just trying to keep it consistent with kubetest2 caller from [kops](https://github.com/kubernetes/kops/pull/18506/changes#diff-448bde7083f881480ae287c451f32bc29e06c7c59c00fb2592063ceaa8ea6d08R239) similar to what we have for [clusterloader2](https://github.com/vishnuchalla/kops/blob/47972e5ac0ec978894d3a9ca89cec2ea767e0eb9/tests/e2e/scenarios/scalability/run-test.sh#L246). But open to other suggestions too.

vishnuchalla:
@mengqiy @upodroid PTAL 👋🏼

pohly:
> Just trying to keep it consistent with kubetest2 caller from kops

Exactly - that [cluster tester](https://github.com/kubernetes/kops/blob/master/tests/e2e/pkg/tester/tester.go) is not maintained in the kubetest2 repo.

aojea:
> cluster tester

echoing Patrick comment, is there any reason if this is going to be used only with kops it can not be maintained there?

are plans to use it outside of kops? if that is the case, who is going to maintain it?

upodroid:
hey, thank you for this change.

I recommend you use the exec tester and close this PR for a couple of reasons:
1. the kubeconfig is always set, and there is exactly 1 cluster in it in CI, kube-burner should be inferring it from the correct lcoations by default
2. the current flags are not really doing anything special that you can't do with exec

Here is a good example: https://github.com/kubernetes/kops/blob/master/tests/e2e/scenarios/metrics-server/run-test.sh

PR #343, opened 2026-06-10
kind: install latest version during CI testing

### LINKED ISSUE #342: outdated kind binary doesn't work on Kubernetes master
https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-kubetest2-kind-build-up-down

needs to use the current release due to kubeadm v1beta3 removal

pohly:
The current version of kind (probably from the test image) is too old at the moment. Even if we update it now, this could happen again. Let's instead always test with the latest.

Fixes: https://github.com/kubernetes-sigs/kubetest2/issues/342

### DIFF
--- a/kubetest2-kind/ci-tests/buildupdown.sh
+++ b/kubetest2-kind/ci-tests/buildupdown.sh
@@ -23,6 +23,15 @@ make install
 make install-deployer-kind
 make install-tester-ginkgo
 
+# Latest kind could have breaking changes, but as we don't lock
+# down Kubernetes or kubeadm either, we have to use the latest,
+# potentially not even released revision.
+#
+# This matches how kind is used elsewhere to test Kubernetes master:
+# - https://github.com/kubernetes/test-infra/blob/e1c19ef211ccf4bb242a59e08efc658b511641e6/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml#L26
+# - https://github.com/kubernetes/test-infra/blob/e1c19ef211ccf4bb242a59e08efc658b511641e6/config/jobs/kubernetes-sigs/kind/kind.yaml#L58
+curl -sSL https://kind.sigs.k8s.io/dl/latest/linux-amd64.tgz | tar xvfz - -C "${GOPATH}/bin"
+
 REPO_ROOT=$(git rev-parse --show-toplevel)
 cd "${GOPATH}/src/k8s.io/kubernetes"
 


pohly:
/assign @aojea 

CI hasn't finished yet, but it's past the new `go install kind@latest`, so at least that part worked. Now we just need to see whether that kind binary got picked up.

pohly:
How and when does kind get updated in the image?

It's too old now. It could be too old again in the future...

pohly:
us-central1-docker.pkg.dev/k8s-staging-test-infra/images/kubekins-e2e:v20260601-2cbf4bdb47-master contains kind v0.31.0, the latest release is 0.32.0 (last week). We could update it.

But there was a time period where only kind master worked with Kubernetes master. Only installing kind@latest would have solved the problem during that time. Therefore I think this PR is the right solution because the job is testing with Kubernetes master.

pohly:
The code didn't actually do what I meant to do: it has to be `kind@main` instead of `kind@latest`. In repos with tagged releases, `latest` is the latest released version.

I still think the rationale makes sense. It matches what E2E jobs in Kubernetes do:
https://github.com/kubernetes/test-infra/blob/e1c19ef211ccf4bb242a59e08efc658b511641e6/config/jobs/kubernetes/sig-testing/kubernetes-kind-presubmits.yaml#L26

Here "latest" is really the latest revision from main, not the latest released version.

An improvement over `go install` in this PR would be to also use `curl -sSL https://kind.sigs.k8s.io/dl/latest/linux-amd64.tgz`. Shall I?

aojea:
> An improvement over `go install` in this PR would be to also use `curl -sSL https://kind.sigs.k8s.io/dl/latest/linux-amd64.tgz`. Shall I?

that is actually how ben has it https://github.com/kubernetes/test-infra/blob/e1c19ef211ccf4bb242a59e08efc658b511641e6/config/jobs/kubernetes-sigs/kind/kind.yaml#L58 , that will be more consistent indeed

pohly:
@aojea: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_kubetest2/343/pull-kubetest2-kind-build-up-down/2064984759729131520 passed with the `curl` approach.

aojea:
/lgtm
/approve

Thanks

kubetest2-kind/ci-tests/buildupdown.sh
```diff
@@ -23,6 +23,15 @@ make install
 make install-deployer-kind
 make install-tester-ginkgo
 
+# Latest kind could have breaking changes, but as we don't lock
```
BenTheElder, line 26: Aside: This could be confused with  Go's `@latest` but we didn't want to break the URLs.

This is a HEAD build. Whereas go's `@latest` is the most recent tagged release.

https://github.com/kubernetes-sigs/kind/blob/d6009ccba8df27aa455e85561b73a0170e60c3d6/site/layouts/index.redirects#L5-L6

PR #344, opened 2026-06-18
app: add --metadata flag to record extra key=value pairs in metadata.json

Jefftree:
Adds a repeatable --metadata key=value flag whose pairs are written into ${ARTIFACTS}/metadata.json alongside the version keys. This lets jobs record per-run metadata (e.g. an experiment variant for a TestGrid column header) without hand-writing the file, which kubetest2 would otherwise overwrite at startup.

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -74,7 +74,7 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 		return err
 	}
 
-	if err := writeVersionToMetadataJSON(d); err != nil {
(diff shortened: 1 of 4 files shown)


aojea:
@upodroid this pull-kubetest2-aws-node-e2e job, do we know who maintain it? it seems is failing because is not configured correctly 

> F0810 10:41:52.765225   29324 aws_runner.go:98] While preparing AWS images: reading userdata file "config/ubuntu2404.yaml", open config/ubuntu2404.yaml: no such file or directory

upodroid:
I'll open a PR to fix it, we bumped to ubuntu2604 and forgot to update our job

upodroid (approved):

PR #346, opened 2026-07-29
Graceful shutdown

KacperPietka:
This PR refactors the kubetest2 runner core signal handling and subprocess execution to use Go's standard context package for cancellation propagation.

Currently, when kubetest2 receives a termination signal (such as SIGINT or SIGTERM sent by a CI environment on timeout), it catches the signal, triggers the deployer's Down() phase, and then exits abruptly with exit code 0 via os.Exit(0).

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/signal"
@@ -87,28 +88,20 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 	}
 	writer := metadata.NewWriter("kubetest2", junitRunner)
 
-	done := make(chan bool)
-	defer func() { done <- true }()
-	go func() {
-		c := make(chan os.Signal, 1)
-		signal.Notify(c, os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
+	normalExit := make(chan struct{})
+	defer func() {
+		close(normalExit)
+		cancel()
+	}()
 
-		// catch interrupt signals and gracefully attempt to clean up
-		for {
-			select {
-			case <-c:
-				if opts.ShouldUp() || opts.ShouldTest() {
-					if opts.ShouldDown() {
-						klog.Info("Captured ^C, gracefully attempting to cleanup resources..")
-						if err := writer.WrapStep("Down", d.Down); err != nil {
-							result = err
-						}
-					}
-					os.Exit(0)
-				}
-			case <-done:
-				return
-			}
+	go func() {
+		<-ctx.Done()
+		select {
+		case <-normalExit:
+			// Normal exit, do nothing
+		default:
+			klog.Info("Captured interrupt, gracefully attempting to cleanup resources..")
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


Qqkyu (commented):
/ok-to-test

KacperPietka:
Hi,
I wanted to flag that I see the failing test terminates with a following error:

`aws_runner.go:98] While preparing AWS images: reading userdata file "config/ubuntu2404.yaml", open config/ubuntu2404.yaml: no such file or directory`

I believe this is not related to my changes but to the test configuration.

aojea:
> Currently, when kubetest2 receives a termination signal (such as SIGINT or SIGTERM sent by a CI environment on timeout), it catches the signal, triggers the deployer's Down() phase, and then exits abruptly with exit code 0 via os.Exit(0).

we have a shutdown that we first tear down all the resources and after that we exits, that looks correct.

With this change we abruplty exit after 10 mins ... I may not understand well this change

KacperPietka:
Thanks for flagging. To clarify, `Down()` still runs after this change. It just now executes after the subprocess exits, instead of racing with it like before.

So, before the change, we abruptly exit. After this change, we don’t (See os.Exit(0) which skips deferred cleanup such as junitRunner)

WaitDelay is the *upper bound* on that graceful shutdown wait: we give the tester up to 10 minutes to exit after the cancellation, after which os/exec forcibly terminates it. So you're right that, in the worst case, this can also delay Down() by up to 10 minutes.

Happy to shorten `WaitDelay` to maybe 5 minutes. Let me know what you think

pkg/app/app.go
```diff
@@ -87,28 +88,20 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
...
-			select {
-			case <-c:
-				if opts.ShouldUp() || opts.ShouldTest() {
-					if opts.ShouldDown() {
-						klog.Info("Captured ^C, gracefully attempting to cleanup resources..")
-						if err := writer.WrapStep("Down", d.Down); err != nil {
-							result = err
-						}
-					}
-					os.Exit(0)
```
aojea, line 107: but how is the Down process triggered now?

KacperPietka: defer func() {
		if opts.ShouldDown() {
			// TODO(bentheelder): instead of keeping the first error, consider
			// a multi-error type
			if err := writer.WrapStep("Down", d.Down); err != nil && result == nil {
				result = err
			}
		}
	}()

pkg/app/app.go
```diff
@@ -87,28 +88,20 @@ func RealMain(opts types.Options, d types.Deployer, tester types.Tester) (result
 	}
 	writer := metadata.NewWriter("kubetest2", junitRunner)
 
-	done := make(chan bool)
-	defer func() { done <- true }()
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
```
aojea, line 91: if the context is cancelled is not also cancelling the current running commands?

KacperPietka: The context cancellation is only relevant to commands created with that context. So the workflow would be:

1. Context is cancelled.
2. Commands created with that context receive the cancellation.
3. this sends interrupt to the tester, giving it a chance to shut down gracefully.
4. WaitDelay puts an upper bound on how long we wait for it to exit.
5. If that bound is reached, the process is terminated.

KacperPietka: then it goes to RealMain and runs the deferred Down()

aojea:
/lgtm
/approve 

Job failure unrelated https://github.com/kubernetes-sigs/kubetest2/issues/347

PR #349, opened 2026-09-08
drop darwin amd64 and replace it with riscv64

upodroid:
I'm dropping darwin amd64 and replacing it with linux riscv64

/cc @dims


dims:
ok in principle. let me know the ci is green please

upodroid:
urgh, unrelated to my change

https://docs.cloud.google.com/kubernetes-engine/docs/concepts/release-channels#no_channel

GKE no longer allows clusters being created without release channels being set.

I'll fix it up

upodroid:
Bump

